PR #8, opened 2024-04-16
Create skeleton code

### LINKED ISSUE #7: Create code skeleton
We need a starting point for work to get started from. Defining the different components of the project which would allow us to get started. While we will not get the interfaces right the first time we can however create a starting point for others to jump off of.

phillebaba:
This adds some skeleton code to get us started on working on the different components.

Fixes #7


internal/store/store.go
```diff
@@ -0,0 +1,11 @@
+package store
+
+import "context"
+
+type Image struct {
+	Reference string
+}
+
+type Storer interface {
```
phillebaba, line 9: Store is a bad name but I could not come up with something else.

internal/replicate/replicate.go
```diff
@@ -0,0 +1,8 @@
+package replicate
+
+import "context"
+
+type Replicator interface {
+	// Replicate copies images from the source registry to the local registry.
+	Replicate(ctx context.Context, ref string) error
```
phillebaba, line 7: We may need to rethink the replicator in the future, if multiple implementations are going to be pulling the image. If this is the case we probably want to pass a reader to avoid duplicating the image pulling logic.

internal/store/store.go
Vad1mo: idea dump ArtifactSourcer

Vad1mo (approved):

PR #10, opened 2024-04-16
Added MADR for decision making following

OneFlyingBanana:
Also added first adr for comparing Skopeo and Crane


docs/decisions/0000-use-markdown-any-decision-records.md
Vad1mo, line 1: rename templates 

- _adr-template-long.md
- _adr-template-short.md

Vad1mo: Remove dir. `node_modules` and the respective package .json files.

Vad1mo (approved):

PR #12, opened 2024-04-16
ADR for registry choice

OneFlyingBanana:
Comparison of Docker vs Zot


Vad1mo:
go with zot for now, to get started with.

PR #20, opened 2024-04-26
File fetcher

OneFlyingBanana:
Reworked main/UI to ask for single input that is automatically categorised as URL or filepath.
File is accessed relatively from working directory


OneFlyingBanana:
http-fetcher changes :
Now using demo.goharbor.io's v2 API to :
1. Retrieve list of tags in repository
2. Retrieve individual digests for each tag
3. Store them as key-value pairs

file-fetcher changes :
Created JSON file with name (tag is optional), digest and repository URL as values
Creates docker pull commands to test usability of data
Works by :
1. Retrieving list of digests from file
2. Retrieving individual image names and tags (if available) for each digest
3. Storing them as key-value pairs

PR #21, opened 2024-05-21
First PoC of Harbor Satellite

OneFlyingBanana:
Features :
- Check user input for local filepath or URL
- Retrieve list of images via URL or local file and store them in-memory
- Check for changes in remote URL or local file and update in-memory store accordingly
- Launch local Zot registry
- Replicate images using in-memory store list to local registry

Main missing features/elements :
- Delete images from local repository if removed from in-memory store
- Handle multiple registries/repositories
- Unit + integration tests
- Optimization and refactoring
- ...


milinddethe15:
Hi @OneFlyingBanana, Can I work on supporting structured logging using log or logrus? 
or adding unit tests for some function?

internal/replicate/replicate.go
```diff
@@ -1,8 +1,133 @@
...
+func (r *BasicReplicator) Replicate(ctx context.Context, image string) error {
+
+	// TODO: Implement deletion of images from the local registry that are not present in the source registry
+	// Probably use crane.Catalog to get a list of images in the local registry and compare to incoming image list
+
+	source := getPullSource(image)
+	fmt.Println("Source:", source)
+
+	if source != "" {
+		CopyImage(source)
```
phillebaba, line 48: Do not ignore the error returned.

internal/replicate/replicate.go
```diff
@@ -1,8 +1,133 @@
...
+	}
+	return nil
+}
+
+func getPullSource(image string) string {
+	input := os.Getenv("USER_INPUT")
+	if os.Getenv("SCHEME") == "https://" {
+		url := os.Getenv("HOST") + "/" + os.Getenv("REGISTRY") + "/" + image
```
phillebaba, line 56: I would avoid getting environment variables here and rather do it at startup and pass then as a parameter. Makes it easier to unit test and debug in the future.

internal/store/http-fetch.go
```diff
@@ -0,0 +1,127 @@
...
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	// Set the Authorization header
+	req.Header.Set("Authorization", "Basic "+auth)
+
+	// Send the request
+	httpClient := &http.Client{}
```
phillebaba, line 53: Create an http client for the struct rather than creating one for every function call.

internal/store/in-memory-store.go
```diff
@@ -0,0 +1,276 @@
...
+	fmt.Printf("Removed image: %s\n", image)
+	return nil
+}
+
+// checkImageAndDigest checks if the image exists in the store and if the digest matches the image reference
+func (s *inMemoryStore) checkImageAndDigest(digest string, image string) bool {
+
+	// Check if the received image exists in the store
+	for storeDigest, storeImage := range s.images {
+		if storeImage == image {
```
phillebaba, line 197: I prefer to inverse the check and call continue instead. This is just to avoid the amount of statement nesting that is done. Improves readability.

internal/replicate/replicate.go
```diff
@@ -1,8 +1,133 @@
...
+	var registryInfo RegistryInfo
+	err = json.Unmarshal(jsonData, &registryInfo)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal JSON: %w", err)
+	}
+
+	return &registryInfo, nil
+}
+
+func CopyImage(imageName string) error {
```
phillebaba, line 100: Does CopyImage need to be exported outside the package.

Vad1mo:
close in favor of the changed code

PR #75, opened 2025-02-09
Refactor `DefaultZotConfig`

mviswanathsai:
While reviewing the satellite code, I realized that DefaultZotConfig is a misleading name. Initially, I mistook it for a constant representing a default configuration for the ZotConfig type.

As we continue building the satellite, we may introduce additional configuration options within DefaultZotConfig to facilitate marshaling config.json. Given this evolution, the name DefaultZotConfig could become increasingly ambiguous.

To improve clarity, I have refactored the structure of ZotConfig (previously DefaultZotConfig) for better readability.

Additionally, I will reword certain parts for further clarity. Would love to hear what you think about this change. We can further refine by adding more configurable fields (that can be un-marshalled), using pointers for optional fields, etc. 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configuration file with updated storage, HTTP, and logging settings.
  
- **Refactor**
  - Improved configuration handling and updated remote registry URL assignments for better consistency.
  - Simplified function return types for error handling.
  - Enhanced context cancellation support to provide more robust shutdown behavior.
  
- **Chores**
  - Removed deprecated configuration options and legacy validation steps.
  - Simplified configuration structure by eliminating outdated extension settings.
  - Updated paths for configuration files to reflect new naming conventions.
  - Removed the old default Zot configuration and related methods.


mviswanathsai:
CC: @bupd @Vad1mo

mviswanathsai:
@Mehul-Kumar-27 I see... scrub seems super useful. I will add the config options in the config struct. And then parse it and update the zot config accordingly. Will that be it, or am I missing something?

Mehul-Kumar-27:
> @Mehul-Kumar-27 I see... scrub seems super useful. I will add the config options in the config struct. And then parse it and update the zot config accordingly. Will that be it, or am I missing something?

Yes, scrub only has one field `enable` just add the configuration so that user can add that, for sync the structure would be quite different. You can look at this for more info how to structure the struct for the sync. https://github.com/project-zot/zot/tree/main/pkg/extensions/sync

mviswanathsai:
Right, thanks for the heads up. As you said, scrub is pretty straight-forward. I will have to take a deeper look at sync.

mviswanathsai:
@Mehul-Kumar-27 it seems like `sync` is used for mirroring purposes. If we allow the zot registry to directly mirror from the upstream repo, what will be the role of satellite in that case?

mviswanathsai:
Should we add a check to satellite to check if an artifact is already present in the local repository before pulling from the upstream to handle these cases?

mviswanathsai:
Right, I wanted to ask about the mention of "full binary" in the zot docs too. Thanks.

We can handle that in another PR if you want. We'll need to check for the extensions field being not nil, and then pull the right binary for zot accordingly.  There might be other hidden nuances to this change though. Do you see anything else?

Vad1mo:
Makes sense

mviswanathsai:
I need to look into the build process for the Zot registry.

mviswanathsai:
I just observed that removing the manual defer to `cancel()` func in `handleRegistrySetup` results in the application not terminating on `ctrl + c`. Thoughts? @Mehul-Kumar-27 @Vad1mo @bupd.

mviswanathsai:
If I remove the `cancel()` func in `handleRegistrySetup()`, and then run `go run main.go` and then hit `ctrl + c`,  the Zot registry shuts down as expected. What I expected was that the execution of main.go should be stopped, but it didn't.

I expected the `ctx.Done()` to receive the signal and thus the server would shut down (inside `SetupServer`). That didn't happen either. Thoughts? @Mehul-Kumar-27 @bupd

mviswanathsai:
> I suggest, we delegate the validation part, to ZOT or alternativly use the ZOT validation on our side

delegated the validation to Zot instead, and removed the data model.

Vad1mo:
@mviswanathsai oing

mviswanathsai:
Rebased and ready to go

mviswanathsai:
Rebased this one. This should be good to review and merge @Vad1mo @bupd

bupd (approved):
LGTM

Vad1mo:
@mviswanathsai, can you fix the linter error?

mviswanathsai:
@Vad1mo, I am actually not getting any linter errors when I run golangci-lint locally. I am unable to locate the issue.

mviswanathsai:
As a sidenote, might be a good idea to bump the `golangci-lint` version that runs in the github workflow. The version `v1.54` of `golangci-lint` we use isn't built with go `v1.24`.

I wonder how it runs without issues. Locally for me, it was complaining until I updated my `golangci-lint` version to `v1.64`.

PR #78, opened 2025-02-12
Fix: Build error

mviswanathsai:
The error is still prevalent, but the test case itself has been refactored.
The source of the e2e test failing is that dagger is not authenticated with docker and docker rate limits dagger from pulling images due to this.

Options:
1. What we can do is, use our own private registry with credentials/ docker account with credentials for testing purposes if we have any. We can configure the test environment with github secret and make use of the credentials as normal env variables.
2. We may need to temporarily remove the test case and keep the issue open until we find a way to authenticate our dagger service. 
3. I tried pulling images from other services (amazon, for example) without authentication but they all require some form of authorization for pulling. We could create a dummy account for this purpose and try if it works.

Need some help here. @Mehul-Kumar-27 @Vad1mo @bupd 

## Summary by CodeRabbit

- **Chores**
  - Consolidated configuration handling by unifying the test configuration file path and removing redundant initialization logic.

- **Tests**
  - Streamlined the test setup by eliminating unnecessary registry setup steps and simplifying image push validations.
  - Enhanced error handling in tests to improve clarity when processing image data.

- **Refactor**
  - Improved internal code clarity by adding descriptive comments to outline future enhancements for image verification.

### DIFF
--- a/internal/state/replicator.go
+++ b/internal/state/replicator.go
@@ -81,6 +81,7 @@ func (r *BasicReplicator) Replicate(ctx context.Context, replicationEntities []E
 	}
 
 	for _, replicationEntity := range replicationEntities {
+        // check if image already exists in zot registry/ local registry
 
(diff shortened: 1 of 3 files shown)


bupd:
@mviswanathsai are you still working on this. can we close this. since builds now work fine.

PR #83, opened 2025-03-11
Remove Cobra Dependency

mviswanathsai:
This PR moves the satellite away from a CLI oriented design that was previously done using the cobra library.


bupd (approved):
lgtm

PR #87, opened 2025-03-11
Remove configurable cronjobs

mviswanathsai:
## Summary
Earlier, the users were allowed to define the job names even though the jobs are mandatory. This PR changes the user flow such that they only define the cron schedule (interval) at which each job will be performed.

Note:
I have still kept the `getters.go` mainly because changing it will result in a cascade effect throughout the code base. 
We will need to do away with the satellite registration cron job and only perform ztr at startup. This will be done in a follow-up PR.

## Summary by CodeRabbit

- **New Features**
  - Introduced new methods to retrieve scheduling intervals more directly.

- **Refactor**
  - Streamlined validation for scheduling intervals to directly check timing expressions and apply defaults when needed.
  - Updated background process scheduling by removing redundant error handling.
  - Removed outdated scheduling configurations to simplify overall operations.


bupd (approved):
lgtm

Issue #79, opened 2025-02-19
Remove the server from the satellite


bupd:
Currently Satellite has both server and client components. 

We need to remove the satellite server components since we are not going to run http-server in satellite.

120EE0980:
@bupd I would like to work on this issue.

Vickysomtee:
@bupd Does this include the server component in the ground-control module?

bupd:
No it's about the server folder on satellite. Nothing to do with ground control

Vickysomtee:
> No it's about the server folder on satellite. Nothing to do with ground control

@bupd for clarity sake, the [root run command](https://github.com/container-registry/harbor-satellite/blob/main/cmd/root.go) in the cmd folder references the server module. Does this mean the run command will also be removed?

psomtee:
> > No it's about the server folder on satellite. Nothing to do with ground control
> 
> [@bupd](https://github.com/bupd) for clarity sake, the [root run command](https://github.com/container-registry/harbor-satellite/blob/main/cmd/root.go) in the cmd folder references the server module. Does this mean the run command will also be removed?

@bupd can you kindly help with an answer to this?

Vickysomtee:
> > No it's about the server folder on satellite. Nothing to do with ground control
> 
> [@bupd](https://github.com/bupd) for clarity sake, the [root run command](https://github.com/container-registry/harbor-satellite/blob/main/cmd/root.go) in the cmd folder references the server module. Does this mean the run command will also be removed?

@bupd Can you kindly help with an answer to this?

PR #94, opened 2025-03-12
fix: remove satellite http server components

narharim:
removed satellite server components

solves #79 

## Summary by CodeRabbit

- **Refactor**
  - Removed web server initialization and related endpoints, shifting operation to core registry management and scheduling.
- **Chore**
  - Eliminated legacy satellite routing functionality, streamlining the application's overall feature set.


mviswanathsai:
Hey @narharim
We discussed yesterday that some of the endpoints are useful. Instead of removing the server, you may want to look into instrumenting metrics with Prometheus. 
We can then decide if we want to keep the server, or adopt a push based approach. 
CC: @Vad1mo @bupd

bupd (changes requested):
Thanks for the contributions. 

Adding request changes.
Since I would like to keep this on hold for sometime until we finalize how do we want to expose metrics in an k8s or k8s like environements. etc.

narharim:
@mviswanathsai Thanks for the suggestion. will start reading about Prometheus integration. can you guide me to resources which would be helpful in instrumenting metrics?

Vad1mo:
@narharim, can you fix the conflicts?

narharim:
@Vad1mo yes, sure. I want to discuss which routes to keep and how we are going to integrate with Prometheus. let's discuss this in the next sync up

Vad1mo:
@narharim, can you resolve the conflicts?

narharim:
@Vad1mo resolved :)

PR #114, opened 2025-04-06
Fix: Infinite replication loop if the replication interval is too short

### LINKED ISSUE #113: BUG: Replication for Big Images being skipped.
For example if an image is size 300mb and it takes around 30 seconds to pull the image on the Network.

But if we set the replication interval to 10s. The satellite replication on large images is getting skipped.

## To-do:
1. Satellite should not start another replication if the previous replication is still running

Thanks

mviswanathsai:
Closes #113 

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced task scheduling to avoid re-running a process that’s already in progress.
	- Adjusted execution flow to ensure tasks proceed consistently without unnecessary pre-checks.


mviswanathsai:
We could also fix this by moving the `defer` logic to after the `isRunning` check. But regardless of that, I think this is a better fix since it lets the scheduler handle when the process should be scheduled.

bupd (approved):
lgtm

PR #129, opened 2025-04-12
Fix satellite registration token generation when no groups are provided

### LINKED ISSUE #108: Satellite Registration without group returns empty array with 200
There is a problem in Ground Control. The problem is that when using `POST /satellites/register` with the following body

```json
{
    "name": "sat1"
}
```

The request with above body fails with Status: 200. and does not have single use token in the response

ambahman:
fixes #108

This fixes the issue where POST /satellites/register with only a name would return 200 but not include a token in the response.

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced reliability by introducing safety checks during data processing to prevent errors when expected information is missing.

- **Refactor**
  - Streamlined error management and adjusted the sequence of operations to ensure that updates occur only when valid data is present, leading to clearer and more robust handling.

### DIFF
--- a/ground-control/internal/server/handlers.go
+++ b/ground-control/internal/server/handlers.go
@@ -343,42 +343,45 @@ func (s *Server) registerSatelliteHandler(w http.ResponseWriter, r *http.Request
 
 	// Give permission to the robot account for the projects present in the group list
 	// fetch all the projects
-	for i := range *req.Groups {
-		projects, err := q.GetProjectsOfGroup(r.Context(), (*req.Groups)[i])
-		if err != nil {
-			log.Println(err)
-			err := &AppError{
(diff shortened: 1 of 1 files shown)


Vad1mo:
The plan is to require as minimum to have a config state. The group membership is not a mandatory requirement.

So the device should be able to register with the config membership only.

bupd (changes requested):
should be handled differently.  I expect satellite will start with a config artifact. so group is not mandatory but config is.

I hope we can put this on hold for now.

bupd:
resolve conflicts and rebase

bupd (approved):
lgtm

Vad1mo:
@amands98 can you resolve the conflicts

PR #118, opened 2025-04-08
Add Dynamic Configuration Endpoints for Ground Control

mviswanathsai:
After this PR, ground control should be able to serve config endpoints, and CRUD configuration states at the upstream harbor registry.

## Summary by CodeRabbit

- **New Features**
  - Added API endpoints for synchronizing, listing, retrieving, deleting, and assigning configurations to satellites.
  - Introduced management of configuration records and their association with satellites, including database and schema support.
  - Enabled creation, update, and deletion of configuration state artifacts in the container registry.

- **Improvements**
  - Integrated configuration selection into satellite registration and management workflows.
  - Enhanced error handling and transaction management for configuration-related operations.

- **Bug Fixes**
  - Limited returned columns in satellite group queries for improved data handling.

- **Chores**
  - Added new dependencies and updated existing ones for registry and scheduling support.

- **Tests**
  - Updated tests to use the newly exported configuration validation function.

PR #132, opened 2025-04-14
added transaction for atomicity and some validations

### LINKED ISSUE #131: transaction issues in addSatelliteToGroup
The current implementation doesn't use transactions and has several edge cases that needs to be handled in graceful manner.

narharim:
Fixes #131 

## Summary by CodeRabbit

- **New Features**
  - Added validation for satellite and group names during grouping requests, providing clear error feedback for invalid inputs.
  - Improved the grouping workflow to recognize when a satellite is already part of a group, avoiding redundant actions.

- **Bug Fixes**
  - Enhanced reliability of satellite-group updates by executing changes within a secure transaction.
  - Improved error handling and logging for satellite-group management operations.


ground-control/internal/server/handlers.go
```diff
@@ -637,7 +637,42 @@ func (s *Server) addSatelliteToGroup(w http.ResponseWriter, r *http.Request) {
...
+		}
+		HandleAppError(w, err)
+		return
+	}
+
+	// Create a new Queries object bound to the transaction
+	q := s.dbQueries.WithTx(tx)
+
+	// Ensure proper transaction handling with defer
+	defer func() {
```
mviswanathsai, line 665: Does this defer logic work? Last I checked, it didn't work as expected for me. i.e, when the handler fails before expected, some entries were still made in the DB. For example, if the "add group" handler fails midway for some reason, the group was still created in the DB. I had to handle it with a different rollback logic. Could you please verify?

ground-control/internal/server/handlers.go
```diff
@@ -658,38 +695,67 @@ func (s *Server) addSatelliteToGroup(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Check if satellite is already in the group
+	groupList, err := q.SatelliteGroupList(r.Context(), sat.ID)
```
mviswanathsai, line 699: Can we directly add a SQL query that returns true or false if a certain entry in this table exists? It would be the equivalent of (sql pseudocode): 
```
SELECT 1 
FROM satellite_groups 
WHERE group_id = ? AND satellite_id = ?;
```

mviswanathsai (commented):
Thanks for your contribution! 
A couple of comments.

ground-control/internal/server/handlers.go
narharim: Thanks for letting me know @mviswanathsai, will verify.

ground-control/internal/server/handlers.go
narharim: We can use the EXISTS operator to return true or false. To perform operations, we need to add a query, and the corresponding code should be generated. Since the rest of the operations are generated by SQLc, I believe we need to generate code for this query using SQLc as well. Right now, I’m not sure if it’s possible to generate code for a specific query using SQLc.

@bupd any thoughts on this?

mviswanathsai: You can add the SQL query in the `queries` directory. Running sqlc will generate the subsequent code. #118 has some relevant code

bupd: @narharim just add your SQL queries to the files inside sqlc/queries directory.

after that run `sqlc generate` to generate the code for the newly added queries

ground-control/internal/server/handlers.go
bupd: it  works for me. It is just for not commiting to the DB.

bupd (changes requested):
I have added my suggestions

pending things are
- create a single query to check if satellite is already in group and return yes or no. instead of going through the list of groups.
- try not to use transactions for the Reads from DB.
- try to simplify and write a clear description of your  changes.

ground-control/internal/server/handlers.go
narharim: @mviswanathsai It's working for me

ground-control/internal/server/handlers.go
narharim: done

bupd (approved):
lgtm

PR #136, opened 2025-04-14
fix: prevent robot account creation on failed registration

### LINKED ISSUE #124: Do Not Create Robot Account if Registration was unsuccessful
Currently if you attempt to create a satellite but the process fails somewhere in the middle, the robot account is still created for the satellite without returning any tokens. The robot account should only be created if the satellite registration is successful.

amands98:
fixes #124

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved resource cleanup to prevent orphaned robot accounts if a transaction fails during satellite registration.

### DIFF
--- a/ground-control/internal/server/satellite_handlers.go
+++ b/ground-control/internal/server/satellite_handlers.go
@@ -84,6 +84,8 @@ func (s *Server) registerSatelliteHandler(w http.ResponseWriter, r *http.Request
 	// Create a new Queries object bound to the transaction
 	q := s.dbQueries.WithTx(tx)
 	committed := false
+	transactionSuccess := false
+	var robotID int64
 
 	// Ensure proper transaction handling with defer
 	defer func() {
@@ -94,6 +96,12 @@ func (s *Server) registerSatelliteHandler(w http.ResponseWriter, r *http.Request
 					Message: "Error: Failed to rollback transaction",
 					Code:    http.StatusInternalServerError,
 				})
+			}
+		}
+		// Cleanup robot account if transaction failed
+		if !transactionSuccess && robotID != 0 {
+			if _, delErr := harbor.DeleteRobotAccount(r.Context(), robotID); delErr != nil {
+				log.Printf("Warning: Failed to cleanup robot account after transaction failure: %v", delErr)
 			}
 		}
 	}()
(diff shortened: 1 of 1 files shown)


bupd (changes requested):
We should not only prevent robot account creation. We should also have a rollback function to delete the robot account if  something goes wrong.

Vad1mo:
The code should also rollback the creation of the robot accoun on harbor side.

ground-control/internal/server/handlers.go
```diff
@@ -369,39 +410,15 @@ func (s *Server) registerSatelliteHandler(w http.ResponseWriter, r *http.Request
...
-	tk, err := q.AddToken(r.Context(), database.AddTokenParams{
-		SatelliteID: satellite.ID,
-		Token:       token,
-	})
-	if err != nil {
-		log.Println("error in token")
-		log.Println(err)
+	transactionSuccess = true
+	if err = tx.Commit(); err != nil {
+		log.Println("Commit failed:", err)
```
bupd, line 417: While testing this when fail to commit to db. robot account is left intact. remove the robot account even on failure to commit
```suggestion
	transactionSuccess = true
	if err = tx.Commit(); err != nil {
		log.Println("Commit failed:", err)
		transactionSuccess = true
```

bupd (approved):
overall lgtm.

Thanks for your contribution @amands98 
only one suggestion. handle fail to commit case.

mviswanathsai:
@amands98 are you working on this? it is very close to completion. I will be ok to take it up and finish it if you don't have the time.

mviswanathsai:
Hope you are feeling better now! 
Just one suggestion: we have moved the code logic into config, groups and satellite handlers. So you will have to move the deletion logic to these new files as the new files also contain some refactoring that we don't want to lose. Please keep this in mind while rebasing

ground-control/internal/server/satellite_handlers.go
```diff
@@ -84,6 +84,8 @@ func (s *Server) registerSatelliteHandler(w http.ResponseWriter, r *http.Request
 	// Create a new Queries object bound to the transaction
 	q := s.dbQueries.WithTx(tx)
 	committed := false
+	transactionSuccess := false
```
mviswanathsai, line 87: Can we just check for `committed` instead? transaction success need not be added on top of that I think.

bupd: good suggestions

bupd (approved):
LGTM

bupd:
closing in favor of #170

PR #141, opened 2025-04-19
feat: add endpoint to delete groups

### LINKED ISSUE #140: feature: Deleting Groups in Harbor Satellite
We need the ability to delete groups in Harbor Satellite. Currently, there’s no functionality to remove a group once it has been created, 

created issue from this discussion -> 
https://github.com/container-registry/harbor-satellite/pull/138#issuecomment-2804662459

amands98:
fixes #140

## Summary by CodeRabbit

- **New Features**
  - Added the ability to delete a group via the API.

- **Bug Fixes**
  - Improved deletion logic for config and satellite state artifacts to ensure correct URL construction and resource cleanup.


bupd (changes requested):
also need  to delete the state artifacts.

bupd (changes requested):
also you need to update all the satellite artifacts. which depends on the deleted group

ground-control/internal/server/handlers.go
```diff
@@ -869,6 +869,183 @@ func (s *Server) getGroupHandler(w http.ResponseWriter, r *http.Request) {
...
+			Message: "Error: Failed to list satellites for group",
+			Code:    http.StatusInternalServerError,
+		}
+		HandleAppError(w, err)
+		tx.Rollback()
+		return
+	}
+
+	for _, satellite := range satellites {
+		if err := s.dbQueries.RemoveSatelliteFromGroup(r.Context(), database.RemoveSatelliteFromGroupParams{
```
bupd, line 923: we are not using the transaction
```suggestion
		if err := q.RemoveSatelliteFromGroup(r.Context(), database.RemoveSatelliteFromGroupParams{
```

ground-control/internal/server/handlers.go
```diff
@@ -869,6 +869,183 @@ func (s *Server) getGroupHandler(w http.ResponseWriter, r *http.Request) {
...
+		if p := recover(); p != nil {
+			tx.Rollback()
+			panic(p)
+		} else if err != nil {
+			tx.Rollback()
+		}
+	}()
+
+	// Remove the group from all associated satellites
+	satellites, err := s.dbQueries.GroupSatelliteList(r.Context(), group.ID)
```
bupd, line 910: also it would be better if we can do the get requests before the transactions.

and have transactions only include the writes

what do you think??

bupd (changes requested):

sujalshah-bit:
@amands98 are you wishing to continue the work? I'd like to contribute :)

PR #142, opened 2025-04-19
fix/log: enforce log level filtering on logger

### LINKED ISSUE #68: Problems in Satellite config
After trying out satellite, I have found a list of errors.

- [ ] Satellite does not respect the `log_level` set in config.json
- [ ] Satellite re-registers everytime eventhough it already has registered. (Satellite registry should be atmost 1)
- [ ] Satellite does not check for updates in ground-control for the satellite state change.
- [ ] 

And will add more.

amands98:
> fixes #68 

> - [ ] Satellite does not respect the log_level set in config.json

Enforce strict log level filtering, so that the log_level set in config.json is respected.

// @bupd @Vad1mo 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved consistency in log level handling to ensure logger output matches the configured log level.

### DIFF
--- a/internal/logger/logger.go
+++ b/internal/logger/logger.go
@@ -56,7 +56,7 @@ func NewLogger(logLevel string) *zerolog.Logger {
 		return fmt.Sprintf("| %s |", l)
 	}
 
-	logger := zerolog.New(output).With().Timestamp().Logger()
+	logger := zerolog.New(output).Level(level).With().Timestamp().Logger()
 
 	return &logger
 }


bupd (changes requested):
Have a look at #144 

we need to have a single logging mechanism. also  should  support  both  json and string format. 

and the actual problem lies with the logs created  from zot registry . which cannot be clearly identified/differentiated from satellite logs. so having a efficient logging mechanism really helps. and this goes even further with having to have a single config for both zot registry  and satellite.

Issue #127, opened 2025-04-12
Improve error handling with a unified and detailed JSON response structure


narharim:
The current AppError structure limits the information sent to clients when errors occur. It only includes a message and a code, which can lead to loss of important context that might be useful for debugging or user feedback.

PR #154, opened 2025-05-12
Add structured error and refactor

vaidikcode:
## Overview
This PR introduces a comprehensive improvement to error handling system to address issue #127. I've replaced the minimal `AppError` structure with a richer, more consistent error response framework that provides better context and actionable information to both users and developers.

### Refactored Components

- Implemented helper functions for common error types:
  - `ValidationError` for input validation failures
  - `NotFoundError` for missing resources
  - `DatabaseError` for database operations
  - `ExternalAPIError` for Harbor API interactions

- Refactored all handlers to use the new error helpers:
  - config_handlers.go
  - group_handlers.go
  - satellite_handlers.go
  - handlers.go
  - helpers.go

## Example

Before:
```json
{
  "message": "Error: Failed to Delete Satellite",
  "code": 500
}
```

After:
```json
{
  "id": "5f3ac342-8671-4d5e-a4bc-e7da1a3c2e35",
  "timestamp": "2023-05-12T14:23:45Z",
  "code": 500,
  "message": "Database operation failed: delete satellite",
  "category": "database",
  "details": "pq: foreign key constraint violation on table 'robot_accounts'",
  "source": "handlers/satellite_handlers.go:224",
  "suggestion": "Ensure satellite is not referenced by other resources before deletion"
}
```

## Summary by CodeRabbit

- **Refactor**
  - Standardized error handling across all server handlers with new error helper functions, providing clearer error categories and more informative messages.
  - Enhanced error responses now include additional metadata such as unique IDs, timestamps, categories, and suggestions for improved clarity.
  - Error reporting is now more consistent and user-friendly throughout the application.
- **Style**
  - Improved consistency in error messages and response formatting for end-users.

### DIFF
--- a/ground-control/internal/server/config_handlers.go
+++ b/ground-control/internal/server/config_handlers.go
@@ -35,11 +35,7 @@ func (s *Server) configsSyncHandler(w http.ResponseWriter, r *http.Request) {
 		if !committed {
 			if err := tx.Rollback(); err != nil {
 				log.Printf("Error: Failed to rollback transaction for failed process: %v", err)
-				err := &AppError{
-					Message: "Error: Failed to rollback transaction",
-					Code:    http.StatusInternalServerError,
-				}
-				HandleAppError(w, err)
+				HandleAppError(w, DatabaseError("rollback transaction", err))
 				return
 			}
 		}
@@ -48,7 +44,7 @@ func (s *Server) configsSyncHandler(w http.ResponseWriter, r *http.Request) {
 	configJson, err := json.Marshal(req.Config)
 	if err != nil {
 		log.Println("Could not marshal JSON: ", err)
-		HandleAppError(w, err)
+		HandleAppError(w, ValidationError("Invalid configuration format", err))
 		return
 	}
 
@@ -67,7 +63,7 @@ func (s *Server) configsSyncHandler(w http.ResponseWriter, r *http.Request) {
 	result, err := q.CreateConfig(r.Context(), params)
 	if err != nil {
 		log.Println("Error persisting config object to database: ", err)
-		HandleAppError(w, err)
+		HandleAppError(w, DatabaseError("create config", err))
 		return
 	}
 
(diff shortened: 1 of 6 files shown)


vaidikcode:
I recognize the potential issues with `runtime.Caller(1)`. I will address that and look into other suggestions.

Vad1mo (changes requested):
I don't think this is the right approach in general. 

This adds complexity, where simplicity is needed, it also increases the error surface. 


I piped my observation into ChatGPT, to elaborate more on why I think It's not right.  I Agree with the observation 90%, so I just copy&paste it here. 

The use of a structured AppError type with fields like ID, Timestamp, Category, Details, and Suggestion adds meaningful context to errors, which is good for observability and debugging.
However, the approach tends to over-engineer error handling by introducing a custom error hierarchy (AppError and its helper methods). This results in increased complexity when simpler, idiomatic patterns could achieve the same goals.
Error Propagation:

Errors are wrapped into AppError multiple times (e.g., ValidationError, NotFoundError, DatabaseError), which adds verbosity and boilerplate code.
This wrapping is not idiomatic in Go, as it makes error propagation verbose and introduces unnecessary abstraction layers.
Coupling Between Error Handling and Business Logic:

Many business logic functions like GetSatelliteByName, ListConfigs, and AddSatelliteToGroup return AppError instances. This tightly couples error handling to business logic, which is contrary to the separation of concerns principle.
Idiomatic Go prefers functions to return plain error values, decoupling error handling from the core logic and allowing for more flexibility.
Overuse of Categories and Metadata:

The error categories (CategoryValidation, CategoryDatabase, etc.) and the Source (file and line number) add some value, but they are not idiomatic Go patterns. Structured logs or observability tools (e.g., OpenTelemetry) are better suited for managing such metadata without complicating the application code.
Handling vs. Logging Errors:

The design conflates error handling and logging by embedding logging logic (e.g., log.Printf) directly into error creation and propagation.
This pattern leads to code duplication and makes testing harder because the error-handling logic is not isolated from the logging and response-writing logic.

vaidikcode:
@Vad1mo i completely agree. I should have discussed on the issue first before making changes. The whole thing just makes the errors more complex. Should i close the PR?

PR #159, opened 2025-05-21
chore: ignore .env and include .env.example for ground-control config.

sujalshah-bit:
Stop tracking `ground-control/.env` and add `.env.example` for environment config.
This helps ensure sensitive or machine-specific configurations aren't committed while still providing guidance for setting up the project.
Should we also apply the same treatment to the root-level .env file?
    

---

## Summary by cubic
Stopped tracking the ground-control/.env file and added a .env.example file for environment configuration.

## Summary by CodeRabbit

- **Chores**
  - Updated the ignore list to exclude environment files in the ground-control directory from version control.


meethereum:
We should also consider ignoring config files perhaps

PR #151, opened 2025-05-04
Dynamic Configuration - Satellite Side Code Update

mviswanathsai:
Enable the satellite to fetch and read the config from upstream. 

## Summary by CodeRabbit

- **New Features**
  - Added file change watcher to detect and respond to configuration updates.
  - Enhanced configuration management with support for previous config tracking and improved validation with default enforcement.
  - Improved logging for configuration warnings and error handling.

- **Bug Fixes**
  - Corrected log messages and improved concurrency safety in process management.

- **Refactor**
  - Simplified process scheduling by removing event broker and cron-based scheduler in favor of a ticker-based approach.
  - Streamlined configuration structures and removed unused or redundant fields and logic.
  - Reorganized container runtime configuration code and improved modularity.

- **Chores**
  - Updated configuration and environment variable handling in deployment files.
  - Improved and refactored test cases to align with updated configuration validation logic.

PR #163, opened 2025-05-26
[WIP] feat: Implement fetching configuration from Ground Control.

sujalshah-bit:
Implemented the logic to fetch configuration from the Ground Control `/configs/{configName}` endpoint. This includes parsing the raw config (JSON) to extract data and emitting a scheduler event to trigger downstream state processing.

Closes TODO in FetchConfigFromGroundControlProcess.Execute().


sujalshah-bit:
This is now obsolete because of #151 I was not aware of this PR Sorry for that 😞

PR #164, opened 2025-05-27
chore: add local config and env files to .gitignore.

sujalshah-bit:
Are there any other files that should be added to .gitignore(like `dagger.json` or `registry_config.json`)? If yes, I’d be happy to include them in this PR as well.


bupd (approved):
lgtm

PR #199, opened 2025-08-14
moved from multi db vars into single db conn string

narharim:
## Summary by cubic
Switched from multiple separate database environment variables to a single database connection string for easier configuration and setup.

- **Refactors**
 - Removed DB_HOST, DB_PORT, DB_DATABASE, DB_USERNAME, and DB_PASSWORD in favor of DB_URL.
 - Updated code to parse and validate the DB_URL variable.

## Summary by CodeRabbit

- Refactor
  - Consolidated DB configuration to a single DB_URL (replacing separate DB_* vars) and updated server to treat PORT as a string with stronger validation and clearer errors.
  - End-to-end and tooling setups now use the unified DB_URL connection string.

- Documentation
  - QUICKSTART and environment example updated to use DB_URL with guidance for local vs. Dagger usage.

- Chores
  - Development service startup now passes the dev runner command as service arguments instead of running it as a separate step.


.dagger/e2e.go
```diff
@@ -18,6 +18,7 @@ const (
 
 	harborAdminUser     = "admin"
 	harborAdminPassword = "Harbor12345"
+	dbConnString        = "postgres://postgres:password@postgres:5432/groundcontrol?sslmode=disable"
```
bupd, line 21: having this for only e2e makes sense. but other areas should always be separate

QUICKSTART.md
```diff
@@ -21,13 +21,8 @@ HARBOR_PASSWORD=Harbor12345
...
 
 PORT=8080
-APP_ENV=local
 
-DB_HOST=127.0.0.1 # For Dagger use DB_HOST=pgservice
-DB_PORT=5432
-DB_DATABASE=groundcontrol
-DB_USERNAME=postgres       
-DB_PASSWORD=password  
+DB_URL="postgres://postgres:password@127.0.0.1:5432/groundcontrol?sslmode=disable" # For Dagger use dbhost=pgservice instead of 127.0.0.1
```
bupd, line 25: This doesn't make sense to have this in QUICKSTART

ground-control/.env.example
```diff
@@ -4,11 +4,5 @@ HARBOR_PASSWORD=
...
 PORT=8080
-APP_ENV=local
 
-DB_HOST=
-DB_PORT=
-DB_DATABASE=
-# Customize user and pass based on your config
-DB_USERNAME=
-DB_PASSWORD=
+DB_URL=
```
bupd, line 8: avoidd using DB_URL. it is not configurable and tougher for users to manage in the long run

ground-control/internal/server/server.go
```diff
@@ -16,33 +17,31 @@ import (
...
 
-	connStr := fmt.Sprintf(
-		"postgres://%s:%s@%s:%s/%s?sslmode=disable",
-		username,
-		password,
-		HOST,
-		PORT,
-		dbName,
-	)
+	connStr := os.Getenv("DB_URL")
```
bupd, line 36: same here

ground-control/internal/server/server.go
```diff
@@ -16,33 +17,31 @@ import (
...
+	}
+
+	_, err := strconv.Atoi(port)
 	if err != nil {
-		log.Fatalf("PORT is not valid: %v", err)
+		log.Fatalf("invalid PORT: %s", port)
 	}
 
-	connStr := fmt.Sprintf(
-		"postgres://%s:%s@%s:%s/%s?sslmode=disable",
```
bupd, line 39: just do remove the sslmode=disable 

add a condition to check for the insecure=true variable in the .env file.

ground-control/internal/server/server.go
```diff
@@ -16,33 +17,31 @@ import (
 )
 
 type Server struct {
-	port      int
+	port      string
```
bupd, line 20: I don't see any added value in having port as string since. we do convert it into integer to check if its a valid port. why change it to string and lose the typecheck

ground-control/internal/server/server.go
```diff
@@ -59,7 +58,7 @@ func NewServer() *http.Server {
 
 	// Declare Server config
 	server := &http.Server{
-		Addr:         fmt.Sprintf(":%d", NewServer.port),
+		Addr:         fmt.Sprintf(":%s", NewServer.port),
```
bupd, line 61: I think this is unnecessary. Keep port as int

ground-control/internal/server/server.go
```diff
@@ -16,33 +17,31 @@ import (
...
-)
-
 func NewServer() *http.Server {
-	port, err := strconv.Atoi(os.Getenv("PORT"))
+	port := os.Getenv("PORT")
+	if port == "" {
+		log.Fatal("PORT is required")
+	}
+
+	_, err := strconv.Atoi(port)
```
bupd, line 31: here

bupd:
closing this as irrelevant. if needed we can open this later

PR #204, opened 2025-08-20
updated quickstart and constants

narharim:


narharim:
@bupd i have tested it and it is working. we just need to have config based on the environment:

* For local development → use **127.0.0.1**
* For running inside a container → use **0.0.0.0**

bupd (approved):
lgtm

PR #329, opened 2026-02-15
feat(spire): enable persistent disk storage for keys

### LINKED ISSUE #288: Embedded SPIRE: Use Persistent KeyManager Instead of In-Memory
## Problem

The embedded SPIRE server in `ground-control/internal/spiffe/embedded_server.go` uses an in-memory KeyManager:

```
KeyManager "memory" {
    plugin_data {}
}
```

This means CA keys regenerate on every restart, invalidating all previously issued SVIDs. Satellites would need to re-register after Ground Control restarts.

## Impact

- All satellite SVIDs become invalid after Ground Control restart
- Satellites must re-register to obtain new SVIDs
- No key continuity across deployments

### Phase 1: Disk-based storage (quick win)

SPIRE natively supports disk-based KeyManager:

```
KeyManager "disk" {
    plugin_data {
        keys_path = "/path/to/keys"
    }
}
```

This requires:
- Persistent volume mount for the keys directory
- Appropriate file permissions (600)

### Phase 2: Database storage (future)

SPIRE's KeyManager is a plugin system. A custom database-backed plugin could:
- Store encrypted keys in PostgreSQL
- Leverage existing DB infrastructure
- Simplify deployment (no extra volumes)

However, this requires writing a custom SPIRE plugin.

### Alternative: Cloud KMS

For cloud deployments, SPIRE supports:
- AWS KMS
- GCP KMS  
- Azure Key Vault

## Reference

- File: `ground-control/internal/spiffe/embedded_server.go:190`
- SPIRE KeyManager docs: https://spiffe.io/docs/latest/deploying/spire_server/

maishivamhoo123:
- Fixes: #288

## Description

This PR enables persistent storage for SPIRE Server keys, resolving an issue where keys were lost upon restart (previously stored in `memory`).

### Changes
- **Updated `KeyManager`:** Switched from `"memory"` to `"disk"` in `internal/spiffe/embedded_server.go`.
- **Configured Storage Path:** Set `keys_path` to `<data_dir>/keys.json`.
- **Added Tests:** Created `internal/spiffe/persistence_test.go` to verify:
  - Correct generation of `server.conf` (checking for `"disk"` and `keys_path`).
  - End-to-End persistence behavior (skips gracefully if `spire-server` binary is missing).

## Additional context
Verified locally with `go test ./internal/spiffe/...`.

---
## Summary by cubic
Enable persistent disk storage for SPIRE Server keys so they survive restarts, replacing the in-memory KeyManager. Addresses #288.

- **Bug Fixes**
  - Set KeyManager to "disk" with keys_path at <data_dir>/keys.json in the embedded server config.
  - Hardened tests: assert disk KeyManager and keys_path in server.conf, add start/stop helpers, stricter E2E (keys.json exists and non-empty), skip if spire-server is missing.

<sup>Written for commit 2609081227f15c12cecd21265dae9548b6aa9105. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Infrastructure**
  * SPIRE server now uses disk-based persistence for configuration and cryptographic keys, with automatic management of persistent data locations.
* **Tests**
  * Added unit and end-to-end tests to validate storage configuration and to verify keys persist and remain usable across server stop/start cycles.
  * Improved test reliability by ensuring HTTP responses are properly closed to prevent resource leaks.


maishivamhoo123:
@bupd can you please review this PR?

ground-control/internal/spiffe/embedded_server.go
```diff
@@ -192,8 +192,10 @@ plugins {
         plugin_data {}
     }
 
-    KeyManager "memory" {
-        plugin_data {}
+    KeyManager "disk" {
+        plugin_data {
+            keys_path = "%s/keys.json"
+        }
```
bupd, line 198: also we should discuss about this first 

does ground-control actually need to store things in disk - whereas it can store them in the DB right ?? - Since ground-control in itself is more of a stateless thing - so I dont want to include the PVC and disk for gc. Hope this makes sense.

bupd (changes requested):
Need to discuss on How GC can persist the keys.

bupd (changes requested):
Core fix is correct — `memory` → `disk` KeyManager addresses #288.

Three issues in `persistence_test.go`:

1. **Missing build tag** — needs `//go:build !nospiffe` at top. Currently breaks `go test -tags nospiffe ./...` since stub has no `writeConfig`.

2. **Broken assertion (L117)** — `currentStat.Size() != initialStat.Size() && initialStat.Size() == 0` never catches the actual failure (both files empty = sizes equal = condition false). Should be `if currentStat.Size() == 0`.

3. **Ignored stat errors (L101, L116)** — `initialStat, _ := os.Stat(...)` can nil-panic on `.Size()` if stat fails. Handle the error.

bupd:

PR #331, opened 2026-02-15
Fix/password policy hash 330

### LINKED ISSUE #330: Password policy is not enforced before hashing in auth module
Currently, in ground-control/internal/auth/password_test.go, passwords are hashed directly using HashPassword() without validating them against the rules defined in ground-control/internal/auth/policy.go.
Example: https://github.com/container-registry/harbor-satellite/blob/main/ground-control/internal/auth/password_test.go

func TestVerifyPassword(t *testing.T) {
    password := "test-password-123"
    hash, err := HashPassword(password)
    require.NoError(t, err)

    tests := []struct {
        name     string
        password string
        hash     string
        want     bool
    }{}
}

However, in other parts of the codebase, password validation and hashing are being handled separately. For example in bootstrap.go:
https://github.com/container-registry/harbor-satellite/blob/main/ground-control/internal/server/bootstrap.go#L32

if err := s.passwordPolicy.Validate(password); err != nil {
    return fmt.Errorf("ADMIN_PASSWORD invalid: %w", err)
}

hash, err := auth.HashPassword(password)

A similar pattern is also repeated in user_handlers.go:
https://github.com/container-registry/harbor-satellite/blob/main/ground-control/internal/server/user_handlers.go#L60

and in token.go:
https://github.com/container-registry/harbor-satellite/blob/main/internal/token/token.go

Instead of validating the password policy separately before calling HashPassword(), we can integrate Validate() inside HashPassword(). This would ensure policy enforcement is consistent everywhere and reduce repeated validation logic across the codebase.

minhpham1810:
- Fixes: #330 

## Description

This PR implements Issue #330 by enforcing the configured password policy directly inside `auth.HashPassword()`.

### Changes

Added password policy validation inside auth.HashPassword()

Introduced `auth.PasswordPolicyError` so callers can treat policy violations as client errors (400) while keeping unexpected hashing failures as server errors (500)

Updated password hashing tests to reflect the new behavior (invalid passwords can no longer be hashed)

### Why

Previously, password validation was performed inconsistently at call sites, which risked missing validation in new code paths. Centralizing validation in `HashPassword()` ensures all password hashes in the system enforce the same policy.

### Testing

`go test ./... in ground-control/`

Fixes #330.

---
## Summary by cubic
Centralized password policy enforcement in auth.HashPassword with PasswordPolicyError so policy violations return 400 and hashing failures return 500. Handlers and bootstrap now rely on this; tests updated; added a Go Report Card badge; fixes #330.

- **Bug Fixes**
  - Enforce env-configured policy in HashPassword; return PasswordPolicyError for violations.
  - Remove duplicate validation in handlers/bootstrap; map policy errors to 400 and other hash errors to 500.

<sup>Written for commit 1788a1f51bab7225764a6e3848e0d1776e421904. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Documentation**
  * Added Go Report Card badge to README.

* **Bug Fixes**
  * Improved password validation and error messaging during user creation and password changes.
  * Streamlined password handling to remove redundant hashing and return clearer client-facing errors for invalid passwords.

* **Tests**
  * Updated password-related tests and test cases to align with validation changes.


ground-control/internal/auth/password_test.go
```diff
@@ -61,28 +61,41 @@ func TestHashPassword(t *testing.T) {
...
 	}{
 		{
 			name:     "normal password",
 			password: "MySecurePassword123!",
+			wantErr:  false,
 		},
 		{
 			name:     "empty password",
 			password: "",
+			wantErr:  true,
```
maishivamhoo123: same as 64.

ground-control/internal/auth/password_test.go
```diff
@@ -61,28 +61,41 @@ func TestHashPassword(t *testing.T) {
...
+			password: "Test@#$%^&*123",
+			wantErr:  false,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			hash, err := HashPassword(tt.password)
+			if tt.wantErr {
+				require.Error(t, err)
```
maishivamhoo123: why ?

minhpham1810:
@bupd @Vad1mo I'd love to receive reviews on this PR. Thank you!

bupd (changes requested):
## Review

The goal of centralizing password policy validation inside \`HashPassword()\` is sound, but there are a few issues worth addressing.

### Critical: Logic ordering bug in \`changeOwnPasswordHandler\`

The PR reorders operations in a way that runs Argon2id hashing **before** verifying the current password:

**Before:**
```
validate policy -> get user -> verify current password -> hash new password -> update
```

**After:**
```
hash new password (argon2id!) -> get user -> verify current password -> update
```

Argon2id is intentionally expensive (19 MiB, 2 iterations per ADR-0004). Any authenticated session can now trigger that computation with a wrong \`current_password\` repeatedly. Fail-fast is violated - the old order was correct.

### Hidden dependency in \`HashPassword()\`

```go
func HashPassword(password string) (string, error) {
    policy := LoadPolicyFromEnv() // reads os.Getenv on every call
    ...
}
```

Previously `s.passwordPolicy` was injected at startup (explicit dependency). Now `HashPassword()` reads env vars as a hidden side effect on every invocation. This makes testing env-dependent (the test comment even acknowledges this: *"flip to true if you change this to exceed policy MaxLength"*). ADR-0004 describes configurable policy via env vars, but the intent was explicit injection, not re-reading env on every hash.

### Policy coupling limits `HashPassword()`

`HashPassword()` is now semantically "validate + hash" rather than just a crypto primitive. Any future call site that needs to hash a pre-validated password (migrations, seeding, admin override) cannot bypass the policy check. Policy enforcement is an application-layer concern; it should not be baked into the crypto primitive.

### Style

- Double blank lines introduced in several places in `user_handlers.go`
- `password.go` uses spaces for the new code block while the rest of the file uses tabs

### Suggestion

The centralization goal is worth pursuing. A cleaner approach would be to keep `HashPassword()` as a pure crypto function and introduce a `ValidateAndHashPassword(policy PasswordPolicy, password string)` wrapper at the application layer - or at minimum fix the operation ordering in `changeOwnPasswordHandler` so current password verification happens before the expensive hash.

The OWASP parameters (Argon2id, 19 MiB, 2 iterations) and the configurable policy env vars are unchanged and remain compliant with ADR-0004.

minhpham1810:
@bupd thank you for the feedback. I will review and make changes.

bupd:

PR #365, opened 2026-03-29
ADR-0008: Air-gapped satellite via virtual twin and signed bundles

Vad1mo:
## Summary

- Proposes air-gapped satellite support via **virtual twins** in Ground Control and **signed OCI Image Layout bundles**
- Virtual twin: same satellite record with `mode = 'airgapped'`, same group/config management — GC generates downloadable bundles instead of expecting live pulls
- Bundle format: `bundle.tar.gz` containing OCI Image Layout, state files (same `PersistedState` format), and Cosign signature
- Trust: configurable OOB (out-of-band key) or TOFU (trust-on-first-use) modes
- Differential bundles: only ship changed images
- Seamless reconciliation: when air-gapped satellite comes online, `state.json` continuity means connected replication picks up where bundles left off
- Zarf analysis: matches Zarf capabilities while adding fleet management and reconciliation

Depends on #364 (ADR renumber). Merge that first.

## Test plan

- [ ] Team review of ADR for architectural alignment
- [x] Validate bundle format against OCI Image Layout spec
- [x] Confirm `go-containerregistry/v1/layout` API matches proposed `LayoutReplicator` design

## Documentation

* Added decision documentation defining end-to-end design for air-gapped satellite deployments using signed portable bundles, new Ground Control API endpoints for bundle and key management, bundle import workflow with signature verification, bundle sequencing to prevent skipped updates, and reconciliation behavior when transitioning between air-gapped and connected deployment modes.


NucleoFusion:
hello @Vad1mo 
I am a little confused to how the reconciliation would work, the document says that reconciliation would:-
```md
- Operator removes --air-gapped, adds --token + --ground-control-url
- Operator does PATCH /api/satellites/{name} {"mode": "connected"} on GC
- Satellite runs ZTR, gets creds, starts normal state replication
```

Now while the 2nd and 3rd points look pretty self-explanatory, does the first point, which is `Operator removes --air-gap and adds --token + --gc-url`. Assuming this Operator is a human and not a daemon or background operation, this feels vague. Does the operator manually stop and restart the satellite? or is there some other way you were considering?

While looking into this, the watcher / config-file handler seems like a good way to address this, since we already add a config entry for the `mode`. We can just have the custom handler for mode changing and do some stuff (which I am looking into more) which would be better than a typical stop and restart, which would require significantly more downtime.

Considering that reconciliation will be one of our flagship features for air-gapped environments, it would make sense to look into this more.

PR #383, opened 2026-05-02
feat: add pagination and filtering to satellite list API

### LINKED ISSUE #382: add pagination and filtering to the satellite list api
## summary

Ground Control currently returns the full satellite list in one response. That is fine for small demos, but it will become harder to use as the number of edge satellites grows.

I think it would be useful to add basic pagination, filtering, and stable sorting to the satellite list api first. This would also help future work on groundctl and the Kubernetes fleet operator, since both will need predictable list behavior for larger fleets.

## problem

Right now clients do not have a clear way to ask for one page of satellites, filter the list, or rely on a deterministic order. For a larger deployment, this can make list views slower and make automation harder to build safely.

For example, a cli command like groundctl satellite list should be able to request a bounded page of results and filter by common operational states instead of always reading the entire fleet.

## proposed direction

Start with the satellite list endpoint and add query parameters such as:

- limit
- offset, or another pagination approach preferred by maintainers
- sort
- order
- filters for fields that already exist, such as activity or stale state, group, config, name prefix, or last seen range

The response could include basic pagination metadata, such as limit, offset, and total count if that is practical with the current database layer.

## initial scope

- satellite list endpoint only
- stable default ordering
- pagination support
- at least one useful filter
- tests for ordering, pagination, and filtering
- keep the current behavior compatible when no query parameters are passed

## out of scope for the first pass

- applying this to users, groups, and configs
- dashboard or ui changes
- changing unrelated api response shapes
- adding label selectors or new satellite metadata

## acceptance criteria

- callers can request a bounded page of satellites
- returned satellites have deterministic ordering
- at least one useful filter is supported
- pagination metadata is returned or documented clearly
- existing callers still work without passing pagination parameters
- tests cover the new query behavior

## related context

This should support the Ground Control cli and Kubernetes fleet operator work being discussed in #375.

bhuvan-somisetty:
Fixes #382

## Description
- Add query controls to GET /api/satellites for limit, offset, name_prefix, sort, and order
- Return pagination metadata when query parameters are used while keeping the existing array response without query parameters
- Add deterministic ordering for the default satellite listing
- Cover pagination, filtering, and validation in handler tests

## Testing
- Not run; Go is not installed in this environment

## Summary by CodeRabbit

* **New Features**
  * Satellite list supports case-insensitive name-prefix filtering (with proper escaping), sorting (allowlisted), and pagination (limit, offset); responses include pagination metadata (limit, offset, total).
  * Results are now deterministically sorted by name then id.

* **Tests**
  * Expanded tests covering filtered listing, pagination, deterministic ordering, and request validation (invalid query params return 400).

### DIFF
--- a/ground-control/internal/database/satellite_list.go
+++ b/ground-control/internal/database/satellite_list.go
@@ -53,7 +53,7 @@ LIMIT $%d OFFSET $%d`, whereSQL, sortColumn, order, limitArg, offsetArg)
 	}
 	defer rows.Close()
 
-	var items []Satellite
+	items := make([]Satellite, 0)
 	for rows.Next() {
 		var i Satellite
 		if err := rows.Scan(
@@ -83,8 +83,8 @@ func satelliteListWhere(arg ListSatellitesFilteredParams) (string, []any) {
 	var args []any
 
 	if arg.NamePrefix != "" {
-		args = append(args, arg.NamePrefix)
-		clauses = append(clauses, fmt.Sprintf("lower(name) LIKE lower($%d) || '%%'", len(args)))
+		args = append(args, escapePostgresLikePattern(arg.NamePrefix))
+		clauses = append(clauses, fmt.Sprintf("lower(name) LIKE lower($%d) || '%%' ESCAPE '\\'", len(args)))
 	}
 
 	if len(clauses) == 0 {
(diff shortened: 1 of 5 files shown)


bupd:

PR #388, opened 2026-05-05
feat: add configurable bandwidth throttling for image replication

### LINKED ISSUE #387: Feature: Configurable Bandwidth Throttling and Sync Scheduling for Satellite Image Replication
## Summary

Satellites are deployed at the edge — on factory floors, retail sites, remote offices — where wide-area network (WAN) links are often shared with operational traffic, metered, or severely constrained. Today, the satellite's image replication process runs with no bandwidth controls: it pulls as fast as the link allows. A large image sync can saturate an edge uplink, disrupting production workloads running on the same device or network segment.

This issue proposes adding **configurable bandwidth throttling** and **sync-window scheduling** so operators can keep image replication within budget without compromising other traffic.

---

## Problem

- Image replication (via `crane`/`skopeo`) is uncapped — a multi-GB base image push can exhaust a metered or constrained edge link.
- There is no way to say "sync only between 02:00–05:00 local time" (e.g., overnight maintenance windows).
- There is no backpressure mechanism: if the link degrades mid-sync, the satellite just retries the whole image, wasting quota.
- Operators have no visibility into bytes consumed per satellite or per sync cycle (related: #63).

---

### 1. Transfer Rate Limit

Add a `max_bandwidth_mbps` configuration field (per-satellite and per-group). Implement a **token-bucket** or **leaky-bucket** rate limiter that wraps the HTTP transport used by `crane`/`skopeo` during image pull/push.

```json
// config.json (satellite)
{
  "sync": {
    "max_bandwidth_mbps": 10,
    "burst_mbps": 20
  }
}
```

The limiter should be applied at the `io.Reader`/`io.Writer` layer so it caps actual bytes-on-wire, not just request count.

### 2. Sync Window Scheduling

Add a `sync_window` stanza that restricts when the satellite is allowed to initiate a new replication cycle. If a cycle is already running when the window closes, it completes gracefully rather than being killed.

```json
{
  "sync": {
    "sync_window": {
      "enabled": true,
      "start": "02:00",
      "end":   "05:00",
      "timezone": "America/Chicago"
    }
  }
}
```

The existing `scheduler` package would be extended to check the window before dispatching `FetchAndReplicateStateProcess`.

### 3. Resume / Partial-Layer Deduplication

When a sync is interrupted (network drop, window close), the satellite should avoid re-pulling layers it already has locally. OCI layer blobs are content-addressable by digest — the local Zot registry already stores them. The replication step should check the local registry's blob store before issuing a pull, skipping layers that are already present.

### 4. Bandwidth Reporting

Extend the heartbeat/status report payload sent to Ground Control to include `bytes_transferred_last_cycle` and `bytes_transferred_total`. This closes part of the gap tracked in #63.

---

## Configuration Reference (proposed)

| Field | Type | Default | Description |
|---|---|---|---|
| `sync.max_bandwidth_mbps` | float | `0` (unlimited) | Peak transfer rate cap in Mbps |
| `sync.burst_mbps` | float | `= max_bandwidth_mbps` | Token-bucket burst allowance |
| `sync.sync_window.enabled` | bool | `false` | Enable time-window restriction |
| `sync.sync_window.start` | string | `"00:00"` | Window open time (HH:MM, 24h) |
| `sync.sync_window.end` | string | `"23:59"` | Window close time |
| `sync.sync_window.timezone` | string | `"UTC"` | IANA timezone identifier |

Ground Control should also be able to **push** these settings to a satellite or satellite group, overriding local config — so operators can enforce fleet-wide policies without SSHing into each node.

---

## Why This Hasn't Been Easy to Add

The current replication path (`internal/state` → `crane` HTTP client) does not expose a hookable transport layer in a clean way. Implementing this requires:

1. Injecting a custom `http.RoundTripper` that wraps `golang.org/x/time/rate` into the crane client.
2. Wiring the rate-limiter config through the satellite config loader and validating it.
3. Modifying the scheduler to be window-aware without breaking the existing interval logic.
4. Extending the heartbeat payload and GC database schema for byte-transfer metrics.

---

## Acceptance Criteria

- [ ] `max_bandwidth_mbps: 10` in config limits satellite replication to ≤ 10 Mbps (verified via `iperf3` or equivalent in an e2e test).
- [ ] With `sync_window` set, the satellite does not initiate a new replication cycle outside the window; an in-progress cycle completes.
- [ ] `max_bandwidth_mbps: 0` (default) disables throttling — no performance regression.
- [ ] Bytes-transferred is included in the satellite heartbeat payload and stored in Ground Control.
- [ ] Ground Control can push bandwidth/window config to a satellite group via the existing config sync endpoint.
- [ ] Unit tests for the rate-limiting transport wrapper.
- [ ] e2e test covering throttled sync.
- [ ] Docs updated (`config.example.json`, website).

---

## Related Issues

- #63 — Track and report upstream how many bytes were transferred
- #232 — Report cache misses to Ground Control
- #285 — Short-lived satellite credentials with heartbeat-based refresh
- #229 — Graceful shutdown with state persistence
- #244 — Pre-load images for offline bootstrap (also targets bandwidth-constrained sites)

---

## Labels

`enhancement`, `component:satellite`, `component:ground-control`, `infrastructure`, `priority:medium`

bhuvan-somisetty:
Satellites are deployed at the edge where WAN links are often shared with operational traffic, metered, or severely constrained. When a large image sync kicks off with no limits, it can saturate the uplink and disrupt production workloads on the same network segment. This adds a `max_bandwidth_mbps` knob to cap download throughput during replication.

The implementation wraps the existing `http.RoundTripper` with a token-bucket rate limiter (`golang.org/x/time/rate`, already a transitive dependency) that limits bytes read from the source registry. Burst is set to 64 KiB or a quarter-second of capacity, whichever is larger, so normal-sized HTTP reads never block on an "n exceeds burst" error. Upload to the local Zot registry (loopback) is intentionally not throttled.

Setting `max_bandwidth_mbps` to `0` (the default) disables throttling entirely with no overhead: the rate-limited transport is not constructed.

Config changes:

```json
"app_config": {
  "sync": {
    "max_bandwidth_mbps": 10
  }
}
```

The throttle also applies in insecure (HTTP) mode since that is the most common satellite configuration. The `buildTransport` method was refactored to layer TLS and throttling independently rather than treating them as a single conditional block.

Closes #387.

Tests added in `internal/state/throttle_test.go`:
- pass-through correctness at high limit
- rate enforcement (250 KB at 1 Mbps takes at least 1s)
- context cancellation propagates through the throttled reader
- end-to-end replication via `NewBasicReplicatorWithTLS` with throttle enabled

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -29,6 +29,9 @@
       "enabled": false,
       "registries": ["docker.io"],
       "runtimes": []
+    },
+    "sync": {
+      "max_bandwidth_mbps": 0
     }
   },
   "zot_config": {
--- a/internal/state/replicator.go
+++ b/internal/state/replicator.go
@@ -150,34 +150,16 @@ func (r *BasicReplicator) buildRemoteOptions(ctx context.Context) ([]name.Option
 func (r *BasicReplicator) replicateOne(ctx context.Context, entity Entity, nameOpts []name.Option, pullOpts []remote.Option, pushOpts []remote.Option) error {
 	log := logger.FromContext(ctx)
 
-	srcRef := fmt.Sprintf("%s/%s/%s:%s", r.sourceRegistry, entity.GetRepository(), entity.GetName(), entity.GetTag())
-	dstRef := fmt.Sprintf("%s/%s/%s:%s", r.remoteRegistryURL, entity.GetRepository(), entity.GetName(), entity.GetTag())
-
-	src, err := name.ParseReference(srcRef, nameOpts...)
+	src, dst, err := r.parseReplicationRefs(entity, nameOpts)
 	if err != nil {
-		return fmt.Errorf("parse source ref %s: %w", srcRef, err)
-	}
-
-	dst, err := name.ParseReference(dstRef, nameOpts...)
-	if err != nil {
-		return fmt.Errorf("parse dest ref %s: %w", dstRef, err)
-	}
-
-	// Lazy fetch: only the manifest is downloaded, no layer data yet
-	desc, err := remote.Get(src, pullOpts...)
-	if err != nil {
-		log.Error().Msgf("Failed to fetch image descriptor: %v", err)
 		return err
 	}
 
-	img, err := desc.Image()
+	ociImage, err := fetchOCIImage(src, pullOpts)
 	if err != nil {
-		log.Error().Msgf("Failed to resolve image: %v", err)
+		log.Error().Msgf("Failed to fetch image: %v", err)
 		return err
 	}
-
-	// Lazy OCI conversion, no data materialized
-	ociImage := mutate.MediaType(img, types.OCIManifestSchema1)
 
 	srcDigest, err := ociImage.Digest()
 	if err != nil {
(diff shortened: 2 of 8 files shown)


bupd:

PR #392, opened 2026-05-05
feat: add pagination, search, and group filtering to satellite list API

### LINKED ISSUE #382: add pagination and filtering to the satellite list api

AdeshDeshmukh:
## Summary

Adds pagination and filtering to `GET /api/satellites` via query parameters.

Fixes #382

### New Query Parameters

| Param | Type | Default | Description |
|-------|------|---------|-------------|
| `page` | int | 1 | Page number (1-indexed) |
| `page_size` | int | 20 | Items per page (max 100) |
| `search` | string | — | Substring match on satellite name (ILIKE) |
| `group` | string | — | Filter by group membership |

### Response Format

```json
{
  "satellites": [{...}, {...}],
  "total": 50,
  "page": 1,
  "page_size": 20
}
```

### Changes

- **SQL**: `ListSatellitesFiltered` + `CountSatellitesFiltered` with optional ILIKE name search, group filter via JOIN, and LIMIT/OFFSET
- **Database**: New params types and query methods following sqlc conventions
- **Handler**: Updated `listSatelliteHandler` to parse query params and return paginated response
- **Backward compatible**: Original `ListSatellites()` preserved

### Verification

- `go test ./...` — 177 tests pass (7 new for pagination/filtering)
- `go vet ./...` — clean
- `go build ./...` — root + ground-control unaffected

### DIFF
--- a/ground-control/internal/database/satellites.sql.go
+++ b/ground-control/internal/database/satellites.sql.go
@@ -185,7 +185,7 @@ func (q *Queries) ListSatellites(ctx context.Context) ([]Satellite, error) {
 const listSatellitesFiltered = `-- name: ListSatellitesFiltered :many
 SELECT id, name, created_at, updated_at, last_seen, heartbeat_interval FROM satellites
 WHERE
-  ($1::text IS NULL OR name ILIKE '%' || $1 || '%')
+  ($1::text IS NULL OR LOWER(name) LIKE LOWER('%' || $1 || '%'))
   AND ($2::text IS NULL OR id IN (
     SELECT sg.satellite_id FROM satellite_groups sg
     JOIN groups g ON g.id = sg.group_id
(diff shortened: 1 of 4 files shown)


bupd:
invalid without discussion

PR #406, opened 2026-05-08
fix: return cached images in satellite status

### LINKED ISSUE #401: bug: status endpoint exposes cached artifacts only as raw IDs
## Summary

`GET /api/satellites/{satellite}/status` currently returns the raw `database.SatelliteStatus` model. Cached image data reported by `/api/satellites/sync` is persisted as `artifact_ids`, but the status endpoint only exposes those raw IDs through the generated sqlc field name (`ArtifactIds`) instead of returning a documented/API-friendly cached artifact shape.

## Code references

- `ground-control/internal/server/satellite_handlers.go`
  - `syncHandler` accepts `cached_images`, stores artifacts, resolves their IDs, and writes those IDs into `InsertSatelliteStatusParams.ArtifactIds`.
  - `getSatelliteStatusHandler` calls `GetLatestSatelliteStatus` and writes the raw database status directly as JSON.
- `ground-control/internal/database/models.go`
  - `SatelliteStatus.ArtifactIds []int32` is a sqlc-generated field with no JSON tag.
- `ground-control/sql/queries/satellite_status.sql`
  - `GetLatestSatelliteStatus` uses `SELECT * FROM satellite_status`, so the status payload is coupled to the DB model.

## Why this matters

This makes the Ground Control status API difficult to consume from the planned OpenAPI-generated client and `groundctl` CLI:

- The response exposes raw internal artifact IDs instead of useful cached image metadata such as reference and size.
- The JSON field name follows the sqlc struct field (`ArtifactIds`) rather than a stable API contract such as `artifact_ids` or `cached_images`.
- Clients that want to display cache/status information must make an additional request to `/api/satellites/{satellite}/images` and manually correlate the data.

## Suggested fix

Introduce an explicit response DTO for `GET /api/satellites/{satellite}/status`, for example:

```go
type SatelliteStatusResponse struct {
    ID                 int32         `json:"id"`
    SatelliteID        int32         `json:"satellite_id"`
    Activity           string        `json:"activity"`
    LatestStateDigest  string        `json:"latest_state_digest,omitempty"`
    LatestConfigDigest string        `json:"latest_config_digest,omitempty"`
    CPUPercent         string        `json:"cpu_percent,omitempty"`
    MemoryUsedBytes    int64         `json:"memory_used_bytes,omitempty"`
    StorageUsedBytes   int64         `json:"storage_used_bytes,omitempty"`
    LastSyncDurationMs int64         `json:"last_sync_duration_ms,omitempty"`
    ImageCount         int32         `json:"image_count,omitempty"`
    CachedImages       []CachedImage `json:"cached_images,omitempty"`
    ReportedAt         time.Time     `json:"reported_at"`
    CreatedAt          time.Time     `json:"created_at"`
}
```

Then have the handler resolve artifact IDs through the existing artifact query path, or document `artifact_ids` explicitly if only IDs are intended.

## Acceptance criteria

- `GET /api/satellites/{satellite}/status` returns an explicit API response type instead of the raw sqlc model.
- Cached artifact information is either returned as `cached_images` or intentionally documented as `artifact_ids`.
- Tests cover the status response when the latest status includes cached image artifacts.
- The eventual OpenAPI spec can describe one stable response schema for this endpoint.

bhuvan-somisetty:
## Summary
- add an explicit satellite status response DTO for the status endpoint
- resolve cached artifact IDs into cached_images entries
- cover the status response shape and cached image data in tests

Fixes #401

## Tests
- GOTELEMETRY=off GOCACHE=C:\tmp\go-build-status-cached-images go test ./...

## Summary by CodeRabbit

* **New Features**
  * Satellite status endpoint now includes cached image data from associated artifacts, providing image references and sizes in the response.

* **Tests**
  * Enhanced test coverage for satellite status endpoint to validate artifact retrieval and verify complete response structure including image details.

### DIFF
--- a/ground-control/internal/server/satellite_handlers.go
+++ b/ground-control/internal/server/satellite_handlers.go
@@ -52,6 +52,22 @@ type SatelliteStatusParams struct {
 	CachedImages        []CachedImage `json:"cached_images,omitempty"`
 }
 
+type SatelliteStatusResponse struct {
+	ID                 int32         `json:"id"`
+	SatelliteID        int32         `json:"satellite_id"`
+	Activity           string        `json:"activity"`
+	LatestStateDigest  string        `json:"latest_state_digest,omitempty"`
+	LatestConfigDigest string        `json:"latest_config_digest,omitempty"`
+	CPUPercent         string        `json:"cpu_percent,omitempty"`
+	MemoryUsedBytes    int64         `json:"memory_used_bytes,omitempty"`
+	StorageUsedBytes   int64         `json:"storage_used_bytes,omitempty"`
+	LastSyncDurationMs int64         `json:"last_sync_duration_ms,omitempty"`
+	ImageCount         int32         `json:"image_count,omitempty"`
+	CachedImages       []CachedImage `json:"cached_images,omitempty"`
+	ReportedAt         time.Time     `json:"reported_at"`
+	CreatedAt          time.Time     `json:"created_at"`
+}
+
 func (s *Server) registerSatelliteHandler(w http.ResponseWriter, r *http.Request) {
 	if s.spiffeProvider != nil || s.spireClient != nil {
 		HandleAppError(w, &AppError{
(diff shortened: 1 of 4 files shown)


ground-control/internal/server/satellite_handlers.go
```diff
@@ -673,7 +689,72 @@ func (s *Server) getSatelliteStatusHandler(w http.ResponseWriter, r *http.Reques
...
-	WriteJSONResponse(w, http.StatusOK, status)
+	resp := satelliteStatusResponse(status, nil)
+	if len(status.ArtifactIds) > 0 {
+		artifacts, err := s.dbQueries.GetLatestArtifacts(r.Context(), sat.ID)
+		if err != nil {
+			log.Printf("Failed to get cached images for status: %v", err)
+			HandleAppError(w, &AppError{Message: "failed to get cached images", Code: http.StatusInternalServerError})
+			return
+		}
+		resp.CachedImages = cachedImagesFromArtifacts(artifacts)
```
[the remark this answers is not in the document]
shri771: This is a valid concern, please resolve this.

ground-control/internal/server/satellite_handlers.go
```diff
@@ -673,7 +689,72 @@ func (s *Server) getSatelliteStatusHandler(w http.ResponseWriter, r *http.Reques
...
+	}
+
+	WriteJSONResponse(w, http.StatusOK, resp)
+}
+
+func satelliteStatusResponse(status database.SatelliteStatus, cachedImages []CachedImage) SatelliteStatusResponse {
```
shri771: This takes a cachedImages `[]CachedImage` param, but the only caller passes nil (line 692) and then overwrites `resp.CachedImages` at line 700 so the parameter never carries a value. It could be better to Drop the param or pass the images in directly.

ground-control/internal/server/satellite_listing_test.go
shri771, line 1: It could be good to add an image_count: 0 / zero-value case, an empty-ArtifactIds case asserting cached_images is absent, and a GetLatestArtifacts-error case covering the new 500 branch.

shri771 (changes requested):
Added a few suggestions, but overall the changes look good.

ground-control/internal/server/satellite_handlers.go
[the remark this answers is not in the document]
bhuvan-somisetty: Good catch — fixed the torn read by adding GetArtifactsByIDs and looking artifacts up by the IDs already sitting in status.ArtifactIds instead of re-querying by satellite ID. That keeps the response anchored to the same status row end-to-end.

ground-control/internal/server/satellite_handlers.go
bhuvan-somisetty: Fair, this param was dead weight — dropped it and pass the images in directly now.

ground-control/internal/server/satellite_listing_test.go
bhuvan-somisetty: Added all three: a zero image_count / empty-artifact_ids case, and a 500 case for the artifacts lookup erroring out. All green locally.

bupd:

PR #411, opened 2026-05-08
fix: read satellite and group from request body in removeSatelliteFromGroup

bhuvan-somisetty:
- Fixes: #

## Description

`DELETE /api/groups/satellite` has no path parameters, but `removeSatelliteFromGroup` was reading `groupName` and `satelliteName` from `mux.Vars(r)`. On a parameterless route `mux.Vars(r)` returns an empty map, so both variables were always empty strings, causing every request to fail immediately with `"Satellite Not Found"`.

The fix aligns the handler with `addSatelliteToGroup`, which already uses the same `SatelliteGroupParams` request body (`{"satellite": "...", "group": "..."}`) for consistent API design.

**Changes:**
- Replace `mux.Vars(r)` path-variable reads with `DecodeRequestBody` into `SatelliteGroupParams`
- Add `utils.IsValidName` guard on the satellite name (same as `addSatelliteToGroup`)
- No route change required; the existing `DELETE /api/groups/satellite` route is correct

**Before:** every `DELETE /api/groups/satellite` call returned 400 `"Satellite Not Found"` regardless of payload.  
**After:** the handler correctly removes the satellite from the group when a valid `{"satellite": "...", "group": "..."}` body is provided.

## Additional context

Identified in a codebase audit of Ground Control API endpoints. The `addSatelliteToGroup` counterpart already uses the body-based approach, so this fix restores symmetry between the two operations.

Signed-off-by: bhuvan-somisetty <sriramvarunkumar@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Added configurable signature verification for container images before replication, supporting "block" (fail on invalid signatures) and "warn" (log warnings but continue) modes.

* **Refactor**
  * Enhanced satellite group management request handling.
  * Refactored replication workflow to integrate signature verification checks.

* **Tests**
  * Added comprehensive test coverage for signature verification policies and replication behavior.

### DIFF
--- a/ground-control/internal/server/satellite_handlers.go
+++ b/ground-control/internal/server/satellite_handlers.go
@@ -1220,9 +1220,23 @@ func (s *Server) addSatelliteToGroup(w http.ResponseWriter, r *http.Request) {
 
 // If the satellite is removed from the group, the state artifact must be updated accordingly as well.
 func (s *Server) removeSatelliteFromGroup(w http.ResponseWriter, r *http.Request) {
-	vars := mux.Vars(r)
-	groupName := vars["group"]
-	satelliteName := vars["satellite"]
+	var req SatelliteGroupParams
+
+	if err := DecodeRequestBody(r, &req); err != nil {
+		HandleAppError(w, err)
+		return
+	}
+
+	if !utils.IsValidName(req.Satellite) {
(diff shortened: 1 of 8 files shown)


bupd:

PR #423, opened 2026-05-10
fix: enforce account lockout for basic auth in authmiddleware

### LINKED ISSUE #422: bug: Basic auth in AuthMiddleware bypasses account lockout mechanism
## Bug

`loginHandler` in `auth_handlers.go`  conducts account lockout correctly — it checks `GetLoginAttempts` before verifying credentials and records failed attempts via `recordFailedAttempt`, and locks the account after `maxFailedAttempts` (5) failures.

`AuthMiddleware` in `middleware.go` accepts Basic auth credentials but does neither:

```go
if username, password, ok := extractBasicAuth(r); ok {
    dbUser, err := s.dbQueries.GetUserByUsername(r.Context(), username)
    if err == nil {
        valid := auth.VerifyPassword(password, dbUser.PasswordHash)
        if valid {
            // authenticated
        }
    }
}
```

no lockout check before attempting verification. no failed attempt recorded on wrong password. attacker can easily brute-force passwords via Basic auth headers against any authenticated endpoint indefinitely — the lockout mechanism is completely bypassed.

## Impact

the account lockout system provides no protection for any endpoint that accepts Basic auth. This makes `maxFailedAttempts` enforcement meaningless for a significant portion of the attack surface.

## proposed fix

In `AuthMiddleware`, before verifying a Basic auth password:
1. Call `GetLoginAttempts` and reject immediately if the account is locked
2. On failed password verification, call `recordFailedAttempt`
3. On successful verification, call `ResetLoginAttempts`

anirudh240:
- Fixes: #422 

## Description
`AuthMiddleware` supported two authentication methods — Bearer token and Basic auth. the bearer path was ok but the basic auth path had no integration with the account lockout system.

`loginHandler` correctly checks lockout status before verifying credentials and it also records every failed attempt. `AuthMiddleware` did none of the above — it called `GetUserByUsername` and `VerifyPassword` directly with no side effects on failure which means an attacker could brute-force credentials via Basic auth against any authenticated endpoint indefinitely, making `maxFailedAttempts` enforcement meaningless for a significant portion of the attack surface.

## Additional context
no new dependencies introduced. Existing `recordFailedAttempt` and DB query methods reused directly

### DIFF
--- a/ground-control/internal/server/middleware.go
+++ b/ground-control/internal/server/middleware.go
@@ -2,9 +2,12 @@ package server
 
 import (
 	"context"
+	"database/sql"
 	"encoding/base64"
+	"errors"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/container-registry/harbor-satellite/ground-control/internal/auth"
 )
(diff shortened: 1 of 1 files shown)


bupd:
please create only one pr as a First-time contributor, Thanks for your contribution

PR #416, opened 2026-05-09
chore: scaffold operator module with controller-runtime and helm v3

bhuvan-somisetty:
Closes #412.

The LFX 2026 Term 2 operator deliverable (#375) needs a foundation to build on. Without a dedicated Go module for the operator, contributors have no defined starting point and will inevitably pick incompatible versions of controller-runtime and the Helm SDK.

This PR creates that foundation:

- `operator/` is a new Go module (`github.com/container-registry/harbor-satellite/operator`) with `sigs.k8s.io/controller-runtime v0.20.4` and `helm.sh/helm/v3 v3.17.3` as direct dependencies.
- `operator/api/v1alpha1/` defines `Satellite` and `SatelliteList` CRD types (`SatelliteSpec`, `SatelliteStatus`) with the standard `DeepCopy*` methods required by the runtime scheme.
- `operator/internal/controller/satellite_controller.go` wires a `SatelliteReconciler` that uses the Helm action SDK (`action.NewInstall` / `action.NewUpgrade`) to install or upgrade the existing satellite Helm chart on each reconcile.
- `operator/cmd/main.go` bootstraps a controller-runtime manager with the fleet scheme registered.
- A `build-operator` task is added to `Taskfile.yml` (backed by `taskfiles/build.yml`) so the binary can be built with `task build-operator`. It is intentionally kept separate from the default `build` task to avoid pulling operator dependencies into the existing CI matrix until the module is further along.

All three binaries (satellite, ground-control, operator) compile cleanly. Existing CI jobs are unaffected.

Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>

## Summary by CodeRabbit

* **New Features**
  * Introduced a Kubernetes operator to manage Satellite resources across fleets with automated Helm-based installation and upgrades.
  * Added a Satellite Custom Resource with fields for ground-control connectivity, credentials, config selection, replica groups, status and heartbeat tracking.

* **Chores**
  * Added build tasks and Go module configuration to support compiling the operator binary and its dependencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/container-registry/harbor-satellite/pull/416)

PR #426, opened 2026-05-11
feat: add Kubernetes finalizer to Satellite CRD for Helm cleanup on delete

### LINKED ISSUE #425: bug: operator leaves orphaned Ground Control registrations when Satellite CRD is deleted
## Problem

The `SatelliteReconciler` in the operator scaffold (PR #416) has no Kubernetes finalizer. When a `Satellite` custom resource is deleted, Kubernetes removes the object immediately without giving the controller a chance to call `DELETE /api/satellites/{name}` on Ground Control.

This leaves the satellite registration in the Ground Control database indefinitely. The satellite still appears in `groundctl satellite list`, consumes a robot account in Harbor, and triggers stale-satellite alerts — even though the deployment no longer exists.

## Root cause

`satellite_controller.go` calls `r.Get(ctx, req.NamespacedName, &sat)` and then `client.IgnoreNotFound(err)` on failure, which means once the object is gone the reconciler simply returns nil and does nothing. There is no deletion path.

Without a finalizer the sequence is:

```
kubectl delete satellite edge-tokyo-01
  → Kubernetes deletes object immediately
  → Reconciler sees NotFound, returns nil
  → Ground Control record: still exists
  → Harbor robot account: still active
```

## Expected behaviour

```
kubectl delete satellite edge-tokyo-01
  → Kubernetes sets DeletionTimestamp, blocks deletion
  → Reconciler sees DeletionTimestamp, calls DELETE /api/satellites/{name}
  → Removes finalizer
  → Kubernetes completes object deletion
```

## Fix

1. Add a named finalizer constant:
   ```go
   const finalizerName = "fleet.harbor-satellite.io/gc-deregister"
   ```

2. In `Reconcile`, check `DeletionTimestamp` first:
   ```go
   if !sat.DeletionTimestamp.IsZero() {
       return r.reconcileDelete(ctx, &sat)
   }
   ```

3. `reconcileDelete` calls `DELETE /api/satellites/{name}` via the generated client, then removes the finalizer via `controllerutil.RemoveFinalizer`.

4. On first reconcile of a new object, inject the finalizer via `controllerutil.AddFinalizer` and patch the object before proceeding with registration or Helm install.

## Secondary concern: Helm install failure after GC registration

The current reconcile loop registers the satellite in Ground Control and then installs the Helm chart. If Helm install fails, the GC registration persists with no corresponding deployment. The finalizer + a deferred rollback call on Helm failure would fix this.

## Affected files

- `operator/internal/controller/satellite_controller.go`

## Labels

`bug`, `component:operator`

bhuvan-somisetty:
Closes #425. Builds on the operator scaffold from #416.

Without a finalizer, deleting a `Satellite` custom resource causes Kubernetes to immediately remove the object. The controller sees `NotFound` on the next reconcile and returns nil, so the Helm release keeps running and any corresponding Ground Control registration persists indefinitely.

This adds a `fleet.harbor-satellite.io/gc-deregister` finalizer to the controller. On first reconcile of a new object the finalizer is injected before any Helm work. When the object is deleted, Kubernetes sets `DeletionTimestamp` and blocks removal until the controller runs `reconcileDelete`, which uninstalls the Helm release via `action.NewUninstall` and then removes the finalizer so Kubernetes can complete the deletion.

The reconcile path is now:

- New object with no finalizer: add finalizer and requeue
- Existing object, no deletion: install or upgrade Helm release as before
- Object marked for deletion: uninstall Helm release, remove finalizer

No new dependencies are needed. `controllerutil.AddFinalizer` and `controllerutil.RemoveFinalizer` are part of `sigs.k8s.io/controller-runtime` which is already a direct dependency. The operator module compiles clean with `go build ./...`.

## Summary by CodeRabbit

* **New Features**
  * Added a new Kubernetes operator for managing Satellite fleet resources.
  * Introduced `Satellite` custom resource with support for ground control integration, credentials management, and registration state tracking.
  * Operator now handles automated Helm-based lifecycle management (install, upgrade, uninstall) of Satellite deployments.
  * Added build support for the operator binary across multiple platforms.

[Review Change Stack](https://app.coderabbit.ai/change-stack/container-registry/harbor-satellite/pull/426)


bupd:

PR #428, opened 2026-05-12
fix: cache group state URLs in memory to survive config hot-reloads

AdeshDeshmukh:
### Problem

`setSatelliteConfig` rebuilds the satellite's group state URLs from scratch on every invocation — once on the initial config assignment, and again on every config hot-reload thereafter. Each rebuild runs `SatelliteGroupList` followed by `GetGroupByID` per group (1+N queries). All the results are immediately discarded after the HTTP response.

A `// TODO: Store the groupStates in memory to survive hot reloads` sat in `config_handlers.go:332` for nearly a year, confirming this was a known gap.

### Fix

Added an in-memory `map[int64][]string` cache on the `Server` struct, keyed by satellite database ID, protected by `sync.RWMutex`. The cache is populated on first access and invalidated at every mutation point where group membership changes.

**Read path** (`config_handlers.go`):
- `setSatelliteConfig` now calls `s.cachedGroupStates()` instead of running an inline 1+N query loop. On cache hit, zero DB queries. On miss, falls through to the existing `getGroupStates()` helper and stores the result.

**Invalidation path** (3 mutation points):
- `addSatelliteToGroup` — invalidated after the state artifact is successfully updated
- `removeSatelliteFromGroup` — same
- `deleteGroupHandler` — invalidated per affected satellite inside the cleanup loop

All invalidations happen *after* the `CreateOrUpdateSatStateArtifact` call succeeds but *before* `tx.Commit()`. If the transaction rolls back, the cache entry simply misses on next read and refetches from the DB — no stale data risk.

### Files changed

| File | Change |
|------|--------|
| `server.go` | +3 fields (`sync` import, `groupStatesCache`, `groupStatesCacheMu`), init in `NewServer()` |
| `helpers.go` | +31 lines: `cachedGroupStates()` method, `invalidateGroupStatesCache()` method |
| `config_handlers.go` | −22 lines (inline loop + TODO removed), +2 lines (single cached call) |
| `satellite_handlers.go` | +4 lines: invalidation in `addSatelliteToGroup` and `removeSatelliteFromGroup` |
| `group_handlers.go` | +2 lines: invalidation per satellite in `deleteGroupHandler` |

### Approach and reasoning

I started by tracing every call site that builds `groupStates` to understand the full picture. There are six places across four files that run some variant of the same query loop. The question was: cache everywhere, or cache strategically?

Caching everywhere felt wrong — handlers like `addSatelliteToGroup` and `deleteGroupHandler` are mutating group membership in the same request, so they *need* fresh data to build the updated state artifact. Caching there would just add a layer of indirection with no upside.

The sweet spot is `setSatelliteConfig`. It's the only handler that runs purely on config assignment and hot-reload — it never modifies groups, it only reads them. Of the six call sites, this one benefits most from a cache because hot-reloads hit it repeatedly without any intervening mutation. The other five keep their inline queries because they're the ones doing the mutating.

For the cache structure itself, I kept it deliberately simple — a plain `map[int64][]string` with `sync.RWMutex`. No generics, no third-party cache library, no TTL expiry, no background cleanup goroutine. The `RateLimiter` in the same codebase already uses this exact pattern, so it's familiar ground for anyone reading the code. Entries are tiny (a handful of strings), invalidation covers every write path, and a stale entry for a deleted satellite just sits unused until the process restarts — not worth a sweeper.

The `int64` key type was a minor call. The DB layer uses `int32`, but some callers (like the delete group loop) have `int64` IDs from earlier casts. Using `int64` avoids narrowing at the invalidation sites and keeps the callsites clean.

### Verification

- [x] `go build ./...` — passes
- [x] `go test ./internal/server/...` — 113 tests pass
- [x] `go vet ./...` — clean
- [x] DCO signed off

## Summary by CodeRabbit

* **Refactor**
  * Enhanced satellite configuration performance by implementing server-side group state caching mechanism. The system now maintains cached group state information, automatically refreshing this cache whenever satellite group memberships are modified, groups are deleted, or satellite configurations are updated. This reduces database query overhead and improves overall system efficiency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/container-registry/harbor-satellite/pull/428)

### DIFF
--- a/ground-control/internal/server/config_handlers.go
+++ b/ground-control/internal/server/config_handlers.go
@@ -318,31 +318,11 @@ func (s *Server) setSatelliteConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	groupList, err := q.SatelliteGroupList(r.Context(), sat.ID)
+	groupStates, err := s.cachedGroupStates(r.Context(), int64(sat.ID), q)
 	if err != nil {
-		log.Printf("Could not get satellite group list: %v", err)
-		err := &AppError{
-			Message: "Error: Failed to Add satellite to config",
-			Code:    http.StatusInternalServerError,
-		}
+		log.Printf("Error: Failed to get group states: %v", err)
 		HandleAppError(w, err)
 		return
-	}
-
-    // TODO: Store the groupStates in memory to survive hot reloads
-	var groupStates []string
-	for _, group := range groupList {
-		grp, err := q.GetGroupByID(r.Context(), group.GroupID)
-		if err != nil {
-			log.Printf("Error: Failed: %v", err)
-			err := &AppError{
-				Message: "Error: Failed to Add satellite to config",
-				Code:    http.StatusInternalServerError,
-			}
-			HandleAppError(w, err)
-			return
-		}
-		groupStates = append(groupStates, utils.AssembleGroupState(grp.GroupName))
 	}
 
 	err = utils.CreateOrUpdateSatStateArtifact(r.Context(), sat.Name, groupStates, req.ConfigName)
--- a/ground-control/internal/server/group_handlers.go
+++ b/ground-control/internal/server/group_handlers.go
@@ -315,6 +315,10 @@ func (s *Server) deleteGroupHandler(w http.ResponseWriter, r *http.Request) {
 
 	committed = true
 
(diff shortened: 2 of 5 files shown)


bupd:
invalid without discussion

PR #429, opened 2026-05-12
fix: validate inputs in ConstructHarborDeleteURL to prevent malformed URLs

### LINKED ISSUE #413: bug: ConstructHarborDeleteURL silently constructs malformed URL on empty inputs
## Bug

`ConstructHarborDeleteURL` in `ground-control/internal/utils/helper.go` accepts `repo` and `repoType` as strings but performs no input validation:

```go
func ConstructHarborDeleteURL(repo string, repoType string) string {
    repositoryName := fmt.Sprintf("%s-state/%s/state", repoType, repo)
    doubleEncodedRepoName := url.QueryEscape(url.QueryEscape(repositoryName))
    return fmt.Sprintf("%s/api/v2.0/projects/satellite/repositories/%s", os.Getenv("HARBOR_URL"), doubleEncodedRepoName)
}
```

If either argument is empty, it silently constructs a malformed URL such as:
https://registry/api/v2.0/projects/satellite/repositories/-state%252F%252Fstate

This URL is then passed directly to `DeleteArtifact` which fires a DELETE request against it. The Harbor API will either return an unexpected error or match the wrong repository entirely.

## Call sites

All three callers pass the value unchecked:
- `satellite_handlers.go:1012` — `ConstructHarborDeleteURL(sat.Name, "satellite")`
- `group_handlers.go:318` — `ConstructHarborDeleteURL(groupName, "group")`
- `config_handlers.go:402` — `ConstructHarborDeleteURL(configName, "config")`

## Fix

Change the signature to return `(string, error)`, validate that `repo` is non-empty and `repoType` is one of the known valid values (`satellite`, `group`, `config`), and update all three call sites to handle the error.

bhuvan-somisetty:
Closes #413.

`ConstructHarborDeleteURL` accepted `repo` and `repoType` without any validation and silently built a broken URL when either was empty or `repoType` was outside the known set. A call like `ConstructHarborDeleteURL("", "satellite")` would produce `/api/v2.0/projects/satellite/repositories/-state%252F%252Fstate` and fire a DELETE against it — either hitting the wrong repository or returning an unexpected Harbor error with no useful signal to the caller.

Changed the signature to return `(string, error)`. It now returns an error if `repo` is empty, or if `repoType` is not one of `satellite`, `group`, or `config`. All three call sites in `satellite_handlers.go`, `group_handlers.go`, and `config_handlers.go` are updated to handle the error before calling `DeleteArtifact`.

Added `helper_test.go` covering valid inputs for all three repo types, empty repo, invalid repo type, empty repo type, and both empty. All tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling for artifact deletion operations with improved input validation. Failures during artifact deletion for configurations, groups, and satellites are now properly detected separately and reported with appropriate error messages, ensuring more reliable state management operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/container-registry/harbor-satellite/pull/429)


bupd:

PR #435, opened 2026-05-14
fix: continue replication on per-image failure instead of aborting group

### LINKED ISSUE #434: bug: single failing image in a group silently blocks replication of all subsequent images - later images never delivered
## What's happening

When the satellite replicates a group of images, `BasicReplicator.Replicate()` processes entities in order and returns immediately on the first error:

```go
// internal/state/replicator.go:119-122
src, err := name.ParseReference(srcRef, nameOpts...)
if err != nil {
    return fmt.Errorf("parse source ref %s: %w", srcRef, err)  // ← entire batch aborted
}

// internal/state/replicator.go:169-172
if err := remote.Write(dst, ociImage, pushOpts...); err != nil {
    log.Error().Msgf("Failed to replicate image: %v", err)
    return err  // ← entire batch aborted
}
```

`processGroupState` treats this as a full failure and returns early — meaning `f.stateMap[index].Entities` is **never updated** (lines 487-489 in `state_process.go` are only reached on success):

```go
// internal/state/state_process.go:470-474
if err := replicator.Replicate(ctx, replicateEntity); err != nil {
    stateFetcherLog.Error().Err(err).Msg("Error replicating state")
    result.Error = fmt.Errorf("failed to replicate entities for %s: %w", ...)
    return result  // ← stateMap.Entities stays nil/stale
}

// never reached on failure:
mutex.Lock()
f.stateMap[index].State = newState
f.stateMap[index].Entities = FetchEntitiesFromState(newState)  // line 488
mutex.Unlock()
```

Because `stateMap[index].Entities` stays as the old value, the next sync cycle sees all the same images as "new" and tries again — hitting the exact same failure — creating a permanent stall.

## Concrete scenario

You're managing 40 edge satellites at factory floors. Each satellite is assigned a group with 8 container images. A developer pushes a new image to Harbor with an accidentally malformed reference (e.g., a repository path containing an invalid character). The group now has this bad image at position #3. Here's what happens on every single sync cycle from that point on:

```
Cycle N:
  Image 1  ✓  already up-to-date, skipped
  Image 2  ✓  already up-to-date, skipped
  Image 3  ✗  ParseReference fails → Replicate() returns error
  Image 4  never reached
  Image 5  never reached
  Image 6  never reached
  Image 7  never reached
  Image 8  never reached

stateMap.Entities unchanged → next cycle is identical

Cycle N+1: same result
Cycle N+2: same result
...forever
```

Images 4–8 are valid, exist in Harbor, and would replicate in seconds — but they are permanently blocked by image 3. The satellite continues heartbeating to Ground Control with no indication of which specific images are stuck. Operators see the satellite as "active" but workloads that need images 4–8 can't pull them locally and either fail to start or silently fall back to the central Harbor — undoing the entire point of the satellite.

## Why this is worse than it appears

1. **Silent starvation** — The heartbeat to Ground Control reports overall group failure but not *which* images are stuck. There's no per-image failure visibility in the status endpoint.

2. **The bad image doesn't have to be new** — An image can become un-pullable after initial replication: Harbor project permissions change, the robot account loses access to a specific project, the image is deleted from Harbor but not from the group. Any of these turn a previously working group into a permanently stuck one.

3. **The staleMap never catches up** — Because `stateMap[index].Entities` is not updated on partial completion, healthy images that were never attempted don't get tracked as "already replicated". They're retried from scratch every single cycle.

4. **Grows worse with group size** — The larger your groups, the more images are silently starved by a single bad actor at the front of the list.

## Expected behavior

The replicator should attempt every entity in the group regardless of individual failures. A per-entity error should be:
1. Logged with the specific image name and reason
2. Collected so the caller gets a meaningful summary of what failed vs what succeeded
3. Not stop healthy images from being replicated
4. Ideally reported back through the heartbeat so Ground Control and operators know *which* images are stuck

## Steps to reproduce

1. Create a satellite group with 5 images in Harbor
2. Manually insert a sixth image entry into the group's `artifacts.json` with a deliberately malformed repository reference
3. Watch the satellite logs — images after the broken one never appear in replication logs
4. Remove the broken entry — images after it now replicate immediately on the next cycle, confirming they were always reachable but were blocked

## Suggested fix direction

Change `Replicate()` to continue on per-image failure and collect errors:

```go
func (r *BasicReplicator) Replicate(ctx context.Context, replicationEntities []Entity) error {
    var failed []string
    for _, entity := range replicationEntities {
        select {
        case <-ctx.Done():
            return ctx.Err()
        default:
        }
        if err := r.replicateOne(ctx, entity, ...); err != nil {
            log.Error().Err(err).Msgf("Failed to replicate %s, continuing with remaining images", entity.GetName())
            failed = append(failed, fmt.Sprintf("%s: %v", entity.GetName(), err))
            continue  // don't abort the batch
        }
    }
    if len(failed) > 0 {
        return fmt.Errorf("%d image(s) failed: %s", len(failed), strings.Join(failed, "; "))
    }
    return nil
}
```

`processGroupState` would then need to distinguish "partial success" from "total failure" so it can still update `stateMap[index].Entities` with the images that succeeded — preventing re-replication of already-healthy images and making progress visible to Ground Control.

## Affected files

- `internal/state/replicator.go` — `Replicate()` is fail-fast on first error
- `internal/state/state_process.go` — `processGroupState()` skips entity tracking update on any error

bhuvan-somisetty:
Closes #434

## What this fixes

When any image in a group failed to replicate, `Replicate()` returned immediately and all subsequent images were skipped. Because `stateMap.Entities` was only updated on full success, the same images were permanently blocked on every future sync cycle.

## Changes

- `Replicate()` in `internal/state/replicator.go` now attempts every entity regardless of individual failures. Failed images are collected and returned alongside a joined error instead of aborting the batch. Per-image logic is extracted into a private `replicateOne()` method.
- `processGroupState()` in `internal/state/state_process.go` updates entity tracking even on partial failure, excluding only the images that actually failed so they are retried next cycle. A new `subtractEntities()` helper computes the succeeded set.
- Context cancellation still exits immediately without updating state.
- `DeleteReplicationEntity`, direct delivery, and the heartbeat/sync flow are unchanged.

## Tests

- Updated all existing `Replicate()` call sites to the new `([]Entity, error)` signature
- Added `TestReplicate_ContinuesAfterOneFailure`: pushes img1 and img3 to source, leaves img2 absent, verifies img1 and img3 are replicated and img2 is returned in the failed list
- Added `TestSubtractEntities` for the new helper

## Summary by CodeRabbit

* **Bug Fixes**
  * Replication now continues processing all entities even when individual entities fail, instead of stopping at the first failure
  * Failed entities are tracked and reported with detailed error information for better diagnostics
  * Improved handling of context cancellation during replication operations

[Review Change Stack](https://app.coderabbit.ai/change-stack/container-registry/harbor-satellite/pull/435)

### DIFF
--- a/internal/state/replicator.go
+++ b/internal/state/replicator.go
@@ -173,8 +173,7 @@ func (r *BasicReplicator) replicateOne(
 		return fmt.Errorf("compute source digest: %w", err)
 	}
 
-	dstDesc, dstErr := remote.Head(dst, pushOpts...)
-	if dstErr == nil && dstDesc.Digest == srcDigest {
+	if r.isUpToDate(dst, srcDigest, pushOpts) {
 		log.Info().Msgf("Image %s already up-to-date at destination, skipping", entity.GetName())
 		return nil
 	}
@@ -193,6 +192,13 @@ func (r *BasicReplicator) replicateOne(
 		return fmt.Errorf("write image %s: %w", entity.GetName(), err)
 	}
 	return nil
+}
+
+// isUpToDate reports whether the destination already holds an image with the
+// same digest as srcDigest, meaning no transfer is needed.
+func (r *BasicReplicator) isUpToDate(dst name.Reference, srcDigest v1.Hash, pushOpts []remote.Option) bool {
+	dstDesc, err := remote.Head(dst, pushOpts...)
+	return err == nil && dstDesc.Digest == srcDigest
 }
 
 // countMissingLayers checks which source layers are absent from the destination
--- a/internal/state/replicator.go
+++ b/internal/state/replicator.go
@@ -154,18 +154,11 @@ func (r *BasicReplicator) replicateOne(
 		return fmt.Errorf("parse dest ref %s: %w", dstRef, err)
 	}
 
-	// Lazy fetch: only the manifest is downloaded, no layer data yet
-	desc, err := remote.Get(src, pullOpts...)
+	// Fetch manifest and resolve to OCI image (no layer data yet).
+	img, err := fetchOCIImage(src, pullOpts)
 	if err != nil {
-		return fmt.Errorf("fetch image descriptor for %s: %w", entity.GetName(), err)
+		return fmt.Errorf("fetch image %s: %w", entity.GetName(), err)
 	}
-
-	img, err := desc.Image()
-	if err != nil {
-		return fmt.Errorf("resolve image %s: %w", entity.GetName(), err)
-	}
-
-	// Lazy OCI conversion, no data materialized
 	ociImage := mutate.MediaType(img, types.OCIManifestSchema1)
 
 	srcDigest, err := ociImage.Digest()
@@ -192,6 +185,20 @@ func (r *BasicReplicator) replicateOne(
 		return fmt.Errorf("write image %s: %w", entity.GetName(), err)
 	}
 	return nil
+}
+
+// fetchOCIImage fetches the manifest from src and resolves it to a v1.Image.
+// Only the manifest is downloaded; layers remain lazy until explicitly read.
+func fetchOCIImage(src name.Reference, pullOpts []remote.Option) (v1.Image, error) {
+	desc, err := remote.Get(src, pullOpts...)
+	if err != nil {
+		return nil, fmt.Errorf("fetch descriptor: %w", err)
+	}
+	img, err := desc.Image()
+	if err != nil {
+		return nil, fmt.Errorf("resolve image: %w", err)
+	}
+	return img, nil
 }
 
 // isUpToDate reports whether the destination already holds an image with the
--- a/internal/state/replicator.go
+++ b/internal/state/replicator.go
@@ -3,6 +3,7 @@ package state
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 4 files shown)


internal/state/replicator.go
shri771, line 1: The loop treats every non-cancellation error as image-specific and continues. But for batch-level failures (registry down, auth, 429) every remaining entity hits the same error, so we make a full set of doomed attempts and emit one joined error per image. Might be worth distinguishing batch-level from image-level errors — e.g. aborting after N consecutive 5xx/401/429. What are your thoughts on this kind of approach?

shri771 (changes requested):
Added a comment.

internal/state/replicator.go
bhuvan-somisetty: Good point, went ahead and added it — tracking consecutive registry-level failures (401/403/429/5xx via transport.Error) and aborting the rest of the batch after 3 in a row, since at that point every remaining image is going to hit the same wall. Per-image errors (like a missing tag) still don't count toward that and replication keeps going as before. Added a test that spins up a source returning 503s to cover it.

bhuvan-somisetty:
Codacy flagged Replicate going over the 50-line limit with the new abort logic — pulled the failure bookkeeping out into its own function, no behavior change.

bupd:

PR #438, opened 2026-05-15
docs(ground-control): add ADMIN_PASSWORD to env example, compose, and quickstart

### LINKED ISSUE #437: docs(ground-control): ADMIN_PASSWORD missing from .env.example and quickstart, blocks first-run startup
## Summary

Following the [token-based ZTR quickstart](https://github.com/container-registry/harbor-satellite/blob/main/deploy/no-spiffe/quickstart.md) to set up Ground Control locally, the server crashes immediately on first run:

```
 Failed to bootstrap system admin: ADMIN_PASSWORD environment variable is required for initial setup
```

## Problem

The `.env.example` doesn't include `ADMIN_PASSWORD`, so the standard setup step `cp .env.example .env` produces a configuration that hard-crashes at [`bootstrap.go:27-29`](https://github.com/container-registry/harbor-satellite/blob/main/ground-control/internal/server/bootstrap.go#L27-L29). The password must also satisfy a policy (8+ characters, at least one uppercase, one lowercase, one number), but this requirement isn't documented anywhere in the setup path.

Two additional issues in the quickstart:

1. **Missing login step** - Steps 4-6 reference `${AUTH_TOKEN}` in their `curl` commands, but there is no prior step showing `POST /login` to obtain it. Even after resolving the password issue, every API call fails with `401 Unauthorized`.

2. **`docker-compose.yml` also missing** - The `groundcontrol` service environment block omits `ADMIN_PASSWORD`, so `docker compose up` hits the same crash.

## Steps to Reproduce

1. Clone the repo and follow the quickstart:
   ```bash
   cd ground-control
   cp .env.example .env
   # Fill in DB and Harbor credentials
   ```
2. Build and run:
   ```bash
   go build -o ../bin/ground-control .
   ../bin/ground-control
   ```
3. Server crashes with:
   ```
   Failed to bootstrap system admin: ADMIN_PASSWORD environment variable is required for initial setup
   ```

## Where `ADMIN_PASSWORD` exists vs. where it's missing

| Location | Has `ADMIN_PASSWORD`? |
|---|---|
| `ground-control/.env.example` | ❌ Missing |
| `ground-control/docker-compose.yml` | ❌ Missing |
| `deploy/no-spiffe/quickstart.md` (Step 1 `.env` example) | ❌ Missing |
| `website/content/docs/quickstart.md` | ✅ Present |
| `docs/decisions/0004-ground-control-authentication.md` | ✅ Mentioned |

The newer website quickstart already includes this variable, the older setup paths are just out of sync.

## Expected Behavior

- `.env.example` should include `ADMIN_PASSWORD` with a comment explaining the password policy
- `docker-compose.yml` should include `ADMIN_PASSWORD` in the groundcontrol service environment
- The quickstart should include `ADMIN_PASSWORD` in its `.env` example and add a login step showing how to obtain `AUTH_TOKEN` before the first API call

## Additional Context

`ADMIN_PASSWORD` is only used on the very first startup, once the `system_admin` user exists in the database, [`BootstrapSystemAdmin`](https://github.com/container-registry/harbor-satellite/blob/main/ground-control/internal/server/bootstrap.go#L22-L24) short-circuits and the variable is never read again. But its absence on that first run is a hard crash with no actionable guidance for the user.

ashimrai123:
Fixes: #437

## Description

While setting up Ground Control locally by following the [token-based ZTR quickstart](deploy/no-spiffe/quickstart.md), I ran into two blockers:

**1. First-run crash** - After `cp .env.example .env` and filling in the DB/Harbor credentials, `ground-control` exits immediately with:
```
Failed to bootstrap system admin: ADMIN_PASSWORD environment variable is required for initial setup
```
I had to trace through [`bootstrap.go:27-33`](ground-control/internal/server/bootstrap.go#L27-L33) and [`policy.go:20-29`](ground-control/internal/auth/policy.go#L20-L29) to figure out what `ADMIN_PASSWORD` needs to look like (8+ chars, uppercase, lowercase, number). The `.env.example` has no mention of it.

**2. No login step** - Once I got past the crash, the quickstart jumps from "verify health" (Step 3) straight into `curl` commands that use `${AUTH_TOKEN}` (Steps 4–6), but never shows how to get that token via `POST /login`. Every API call returns `401`.

The newer [`website/content/docs/quickstart.md`](website/content/docs/quickstart.md) already has `ADMIN_PASSWORD` and a login step, the older setup docs are just out of sync.

### `ground-control/.env.example`
- Added `ADMIN_PASSWORD=` with a comment explaining the password policy requirements

### `ground-control/docker-compose.yml`
- Added `ADMIN_PASSWORD` to the `groundcontrol` service environment (defaults to `ChangeMe123` via `${ADMIN_PASSWORD:-ChangeMe123}`)

### `deploy/no-spiffe/quickstart.md`
- Added `ADMIN_PASSWORD=Admin1234` to the Step 1 `.env` example
- Inserted new **Step 4: Login to Ground Control** - shows how to obtain `AUTH_TOKEN` via `POST /login` before the first API call
- Renumbered subsequent steps (old 4→5, 5→6, 6→7, 7→8)

## Summary by CodeRabbit

* **Documentation**
  * Added prerequisite for a JSON parser and a new step to obtain an API bearer token (curl + jq), plus renumbered tutorial steps and updated startup instructions to reference the token.
* **Chores**
  * Required an admin password for initial setup with complexity guidance in the example configuration.
  * Made orchestration configuration enforce providing the admin password at runtime.

[Review Change Stack](https://app.coderabbit.ai/change-stack/container-registry/harbor-satellite/pull/438?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/deploy/no-spiffe/quickstart.md
+++ b/deploy/no-spiffe/quickstart.md
@@ -9,6 +9,7 @@ For production deployments with cryptographic identity and mTLS, see the [SPIFFE
 - A Harbor registry instance with the satellite adapter installed. You can find the instance [here](https://github.com/container-registry/harbor-next/tree/satellite).
 - Credentials with permission to create robot accounts in the registry
 - [Task](https://taskfile.dev/installation/) installed.
+- [jq](https://jqlang.github.io/jq/download/) installed (used to parse JSON responses).
 - (Optional) Docker and Docker Compose. [Install Docker](https://docs.docker.com/get-docker/).
 
 ## Step 1: Configure Ground Control
@@ -46,6 +47,10 @@ Ground Control is the central service that manages satellite configurations.
    DB_DATABASE=groundcontrol
    DB_USERNAME=postgres
    DB_PASSWORD=password
+
+   # Ground Control admin (created on first run)
+   # Must be 8+ chars with uppercase, lowercase, and number
+   ADMIN_PASSWORD=Admin1234
    ```
 
    > Note: Ensure the database is running and accessible.
@@ -90,7 +95,23 @@ curl http://localhost:8080/health
 
 A `200 OK` response indicates Ground Control is healthy.
 
-## Step 4: Create a Group for Artifacts
+## Step 4: Login to Ground Control
+
+All `/api/*` endpoints require a Bearer token. Obtain one by logging in with the
+admin account you configured via `ADMIN_PASSWORD`:
+
+```bash
+AUTH_TOKEN=$(curl -s -X POST http://localhost:8080/login \
+  -H "Content-Type: application/json" \
+  -d '{"username": "admin", "password": "Admin1234"}' | jq -r '.token')
+
+echo $AUTH_TOKEN
+```
+
+> Note: Replace the password with whatever you set as `ADMIN_PASSWORD` in your `.env`.
+> The token is valid for 24 hours by default.
+
+## Step 5: Create a Group for Artifacts
 
 A group is a set of images that the satellite needs to replicate from the upstream registry. It also contains information about all the artifacts present in it.
 
@@ -117,7 +138,7 @@ curl -X POST http://localhost:8080/api/groups/sync \
 
 > Note: Replace `repository`, `tag`, and `digest` with your artifact details. Use `docker inspect` or Harbor's UI to find the digest.
 
-## Step 5: Configure the Satellite
+## Step 6: Configure the Satellite
 
 Create a config artifact for the satellite. See [config example](https://github.com/container-registry/harbor-satellite/blob/main/examples/config.json). This artifact tells the satellite where Ground Control is located and defines how and when to replicate artifacts. It also includes the local OCI-compliant registry configuration.
 
@@ -161,7 +182,7 @@ curl -i --location 'http://localhost:8080/api/configs' \
 
 > Tip: Adjust `ground_control_url` and `local_registry.url` if running on a different host or port.
 
-## Step 6: Register the Satellite
+## Step 7: Register the Satellite
 
 Register the satellite with the group and configuration created earlier. This request returns a token. Save it for the next step.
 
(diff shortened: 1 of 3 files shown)


ground-control/docker-compose.yml
```diff
@@ -31,6 +31,7 @@ services:
       HARBOR_PASSWORD: Harbor12345
       HARBOR_URL: http://harbor-sat.kube.bupd.xyz
       PORT: 8080
+      ADMIN_PASSWORD: "${ADMIN_PASSWORD:-Admin1234}"
```
shri771: We should not set the default password,  because this could create a know admin password if the user fails to set it in an env var. Instead we should require this var `${ADMIN_PASSWORD:?ADMIN_PASSWORD must be set}`

deploy/no-spiffe/quickstart.md
```diff
@@ -90,7 +95,23 @@ curl http://localhost:8080/health
...
-## Step 4: Create a Group for Artifacts
+## Step 4: Login to Ground Control
+
+All `/api/*` endpoints require a Bearer token. Obtain one by logging in with the
+admin account you configured via `ADMIN_PASSWORD`:
+
+```bash
+AUTH_TOKEN=$(curl -s -X POST http://localhost:8080/login \
+  -H "Content-Type: application/json" \
+  -d '{"username": "admin", "password": "Admin1234"}' | jq -r '.token')
```
shri771, line 106: nit: we could update the password to a placeholder `<your-admin-password>` instead of hardcoding it.

shri771 (changes requested):
Added a few suggestions, but overall the changes look good.

ground-control/docker-compose.yml
ashimrai123: Thanks for pointing this out. 
Initially, I kept the fallback to align with other connection secrets like DB_PASSWORD and HARBOR_PASSWORD. However, looking deeper into the bootstrap logic, ADMIN_PASSWORD is a one-time setup credential that gets hashed and saved directly to Postgres, making it painful to change later. Enforcing the environment variable at startup if no .env has it is definitely the right call for user convenience. Updated.

deploy/no-spiffe/quickstart.md
ashimrai123: The curl command uses the same Admin1234 shown in the example .env setup in Step 1 (line 53) so the guide remains copy-pasteable for a quick setup. Line 111 already notes to change it: "Replace the password with whatever you set as ADMIN_PASSWORD in your .env."

shri771: Makes sense, thanks for the clarification.

shri771 (approved):
LGTM

bupd:
invalid and has conflicts

PR #448, opened 2026-05-17
feat: add audit logging for security events with syslog and OpenTelemetry transports

### LINKED ISSUE #238: Audit Logging for Security Events
## Context

Security-relevant events should be logged to a structured audit log for compliance and incident investigation.

## Current State

- Application uses zerolog for general logging
- No separate audit log stream
- Security events mixed with operational logs
- No structured format for security events

## Expected Behavior

Separate audit logger with:
- Structured JSON format
- Configurable output file path
- Log rotation support
- Unique event IDs for correlation

Events to audit:
- Authentication failures (user and satellite)
- Satellite registration/deregistration
- Satellite revocation/unrevocation
- Policy violations (blocked pulls)
- Configuration changes
- User management (create/delete/password change)

Example audit log entry:
```json
{
  "event_id": "uuid",
  "timestamp": "2025-01-29T10:00:00Z",
  "event_type": "satellite.auth.failure",
  "actor": "sat-edge-01",
  "details": {"reason": "invalid_token"},
  "source_ip": "10.0.0.5"
}
```

## Value

- Compliance with security frameworks (SOC2, ISO27001)
- Incident investigation capability
- Security monitoring integration

## Changes to be Made

- [ ] Create audit logger package
- [ ] Define audit event types and structure
- [ ] Add audit config (file path, rotation settings)
- [ ] Instrument GC handlers with audit logging
- [ ] Instrument satellite for local audit logging
- [ ] Document audit log format

<details>
<summary>Files to Modify</summary>

Satellite:
- `internal/logger/audit.go` (new)
- `pkg/config/config.go`

Ground Control:
- `ground-control/internal/logger/audit.go` (new)
- `ground-control/internal/server/middleware.go`
- `ground-control/internal/server/satellite_handlers.go`
- `ground-control/internal/server/user_handlers.go`

</details>

Aloui-Ikram:
Closes #238.

## Summary

Adds a dedicated, structured **security audit log** on both **Ground Control** and **Satellite**. Each event is built once into a canonical JSON `Record` and fanned out through a **transport seam**, so a new destination plugs in without touching the event model or the emit path. Two transports ship today (usable together): **syslog (RFC 5424)** and **OpenTelemetry (OTLP/HTTP)**. Disabled by default; `json.Marshal(Record)` is the single wire payload, so transports can never disagree.

```mermaid
flowchart LR
  GC["Ground Control"] --> SEAM["Transport seam<br/>1 canonical JSON Record"]
  SAT["Satellite"] --> SEAM
  SEAM --> SYS["syslog (RFC 5424)"]
  SEAM --> OT["OpenTelemetry (OTLP/HTTP)"]
  SYS --> ST["target: daemon | network | file"]
  OT --> COL["OTel collector"] --> IDX["SIEM (parsed fields)"]
```

## Event model

One canonical JSON record: **8 always-present + 9 optional** fields. syslog wraps it as `<PRI>1 TIMESTAMP HOST harbor-audit PID - - {payload}`; OTel carries `{payload}` as the log `body` with the fields mapped to attributes.

```
<132>1 2026-06-03T10:00:00Z gc-1 harbor-audit 4242 - - {"event_id":"...","severity":"warning","component":"ground-control","event_type":"session.login.failure",...}
```

<details><summary><b>Full field reference (17 fields + OTel mapping)</b></summary>

| Field | Always? | Notes |
|---|---|---|
| `event_id` | yes | UUID v4, per event |
| `timestamp` | yes | UTC, RFC 3339 with nanoseconds |
| `severity` | yes | `info`/`warning`/`error`/`critical`; derived from `outcome`, overridable. -> syslog PRI, OTel `SeverityNumber` |
| `component` | yes | `satellite` or `ground-control`; carried on the record. -> OTel `service.name` |
| `event_type` | yes | Derived `{resource_type}.{operation}.{outcome}`. -> OTel `event.name` |
| `operation` | yes | `login`/`create`/`delete`/`update`/`register`/`deregister`/`password_change`/`auth` |
| `resource_type` | yes | `user`/`satellite`/`config`/`session` |
| `outcome` | yes | `success`/`failure` |
| `actor` | no | Username / satellite name. -> OTel `user.name` |
| `actor_type` | no | `user`/`robot`/`satellite`/`anonymous`/`system` |
| `source_ip` | no | GC client IP (RemoteAddr; `X-Forwarded-For` only if `AUDIT_TRUST_FORWARDED_HEADERS=true`). -> OTel `client.address` |
| `user_agent` | no | Client `User-Agent`. -> OTel `user_agent.original` |
| `request_id` | no | Correlation ID shared by all events from one GC request |
| `satellite_id` | no | The satellite a satellite-scoped event relates to |
| `resource` | no | Concrete target instance (created username, changed config name) |
| `reason` | no | Low-cardinality failure code. -> OTel `error.type` |
| `details` | no | Event-specific map; config changes carry `details.changed.{path}.{from,to}` (secrets redacted). -> OTel `harbor.audit.details` |

Attributes without a semconv name use the `harbor.audit.*` namespace (`operation`, `resource_type`, `outcome`, `actor_type`, `resource`, `request_id`, `satellite_id`, `details`).

</details>

## Event catalogue

`event_type` = `{resource_type}.{operation}.{outcome}`. `satellite.auth.failure` is kept distinct from `satellite.register.failure` so auth-failure alerting is not polluted by benign network errors.

| `event_type` | Source | `reason` values |
|---|---|---|
| `session.login.success` / `session.login.failure` | GC | `missing_credentials`, `account_locked`, `unknown_user`, `bad_password` |
| `user.create.success` / `user.delete.success` / `user.password_change.success` | GC | - |
| `satellite.register.success` | GC + Satellite | - |
| `satellite.register.failure` | Satellite | `registration_failed`, `invalid_state_auth_config` |
| `satellite.deregister.success` | GC | - |
| `satellite.auth.failure` | GC | `invalid_token`, `token_expired`, `missing_spiffe_identity` |
| `config.create.success` / `config.update.success` / `config.delete.success` | GC + Satellite | - |

## Implementation

One call - `Log(AuditEvent)` - builds the event once and fans it out:

```mermaid
flowchart TD
  A["Log(AuditEvent)<br/>no-op if disabled / nil"] --> B["build Record once<br/>(event_id, timestamp, component,<br/>event_type, severity)"]
  B --> C["json.Marshal(Record)<br/>= single wire payload"]
  C --> D{"fan out to transports<br/>(outside the lock)"}
  D --> E["syslog: format RFC 5424<br/>-> daemon | network | file"]
  D --> F["OTel: OTLP/JSON<br/>-> POST /v1/logs"]
```

- **Transport seam.** New destination = one `Transport` interface (`Emit(Record) error`, `Close() error`) registered in `Reconfigure`; emit path unchanged. Delivery is outside the logger lock (a slow sink can't block a hot reload); a transport error is logged, not dropped.
- **Syslog.** Formats RFC 5424 directly (stdlib `log/syslog` only does legacy BSD and can't write a file). `severity` -> priority; rotation only on `file`; destination validated at startup (fails fast).
- **OpenTelemetry.** Hand-rolled OTLP/JSON over `net/http`, no SDK dependency. `component` -> `service.name`, scope `harbor-satellite/audit`, `severity` -> SeverityNumber, fields -> semconv + `harbor.audit.*`, `body` = canonical JSON. Endpoint TCP-probed at startup.
- **Emitters.** `Log` is safe to call unconditionally. On GC, `RequestIDMiddleware` + an `auditEvent` helper fill `source_ip`, `user_agent`, `request_id`, so handlers pass only the semantic fields.

| Area | Files |
|---|---|
| Logger (both modules) | `internal/logger/{audit,syslog,otel}.go` + `ground-control/internal/logger/...` and tests |
| Config | `pkg/config/*`, `config.example.json`, `ground-control/internal/server/server.go`, `.env.example` |
| Wiring | `cmd/main.go`, `ground-control/internal/server/{server,middleware,routes}.go`, `internal/state/registration_process.go` |
| Emitters (GC) | `ground-control/internal/server/{auth,user,satellite,config}_handlers.go` |
| Docs | `docs/guides/audit-logging.md` |

## Configuration

Disabled by default. Satellite uses an `app_config.audit` block; Ground Control uses env vars. Both transports can run together.

```json
"audit": {
  "enabled": true,
  "syslog": { "target": "file", "tag": "harbor-audit", "socket_path": "/dev/log",
    "network": "udp", "address": "siem.example:514",
    "file": { "path": "/var/log/harbor-satellite/audit.log", "max_size_mb": 100, "max_backups": 7, "max_age_days": 30, "compress": true } },
  "otel": { "enabled": true, "endpoint": "http://otel-collector:4318" }
}
```

<details><summary><b>Ground Control env vars</b></summary>

```
AUDIT_LOG_ENABLED=true
AUDIT_SYSLOG_ENABLED=true            # default true; false runs OTel only
AUDIT_SYSLOG_TARGET=file            # daemon | network | file
AUDIT_SYSLOG_TAG=harbor-audit
AUDIT_SYSLOG_SOCKET_PATH=/dev/log   # daemon
AUDIT_SYSLOG_NETWORK=udp            # network: udp | tcp
AUDIT_SYSLOG_ADDRESS=siem.example:514
AUDIT_SYSLOG_FILE_PATH=/var/log/ground-control/audit.log   # file (+ MAX_SIZE_MB / MAX_BACKUPS / MAX_AGE_DAYS / COMPRESS)
AUDIT_OTEL_ENDPOINT=http://otel-collector:4318             # non-empty enables OTel
AUDIT_TRUST_FORWARDED_HEADERS=false
```
</details>

Only the chosen syslog `target`'s settings are read; invalid input makes the process refuse to start rather than silently drop events.

## Verification

`go test ./...` passes in both modules (event shape, severity mappings, the 3 syslog targets incl. RFC 5424 framing + UDP path, the OTLP envelope + attribute mapping, fail-fast on bad target/endpoint, config defaulting/validation).

**End-to-end** on a real stack (Harbor v2.11.1 + GC + Postgres, 2026-06-14): the full reproducible GC catalogue (**10 event types**) was fired against each transport in turn, plus a satellite registered via ZTR and hot-reloaded for the `component: satellite` events. (`satellite.deregister.success` is not locally reproducible - see Known limitations.)

**`config.update` redacted diff** (every transport) - a rotated secret shows as changed without its value ever being logged:

```json
{ "changed": {
    "app_config.ground_control_url": { "from": "http://gc.example:8080", "to": "http://gc.example:9090" },
    "app_config.log_level":          { "from": "info", "to": "debug" },
    "state_config.auth.password":    { "from": "[REDACTED]", "to": "[REDACTED]" } } }
```

**syslog `file`** - RFC 5424 framing on disk, one line per event:

```
<132>1 2026-06-14T13:28:23Z hp harbor-audit 64632 - - {"event_id":"c05f...","component":"ground-control","event_type":"session.login.failure","outcome":"failure","severity":"warning","actor":"admin","source_ip":"::1","reason":"bad_password"}
```

**syslog `daemon`** - all 10 types in journald. journald ignores the RFC 5424 APP-NAME, so `-t` does not match; query the body (or use `rsyslog` / `network`):

```
$ journalctl -t harbor-audit
-- No entries --
$ journalctl --since today -g "harbor-audit" -o cat | jq -r '.event_type+" ["+.severity+"/"+.outcome+"]"'
session.login.failure [warning/failure]
session.login.success [info/success]
user.create.success [info/success]          ... (10 distinct types)
satellite.register.success [info/success]
satellite.auth.failure [warning/failure]
```

**syslog `network` into a vanilla Wazuh (no custom decoder)** - events arrive but are an opaque blob; `wazuh-logtest` on a login failure:

```
**Phase 2: Completed decoding.        No decoder matched.
**Phase 3: Completed filtering (rules).
        Rule id: '1002'   level: '2'   description: 'Unknown problem somewhere in the system.'
```

>Wazuh Discover: `full_log` holds the JSON blob, `decoder name` empty on every row.
03-syslog-network-blob

**Same syslog feed into a second SIEM (vanilla Splunk)** - identical result: the JSON payload lands as an opaque blob, with only Splunk's own metadata extracted, not the audit fields. This confirms the behavior is not Wazuh-specific: a raw syslog payload is SIEM-independent until the operator adds field extraction.

> Splunk: events present, but audit fields are not extracted (left sidebar shows only Splunk defaults).
08-splunk-syslog-blob

One extraction step (`rex` + `spath`, lighter than writing a Wazuh decoder) makes all 17 fields searchable:

> Splunk: parsed table after one extraction step (`event_type` / `outcome` / `actor` / `reason` / `source_ip`).

09-splunk-syslog-parsed

**OpenTelemetry into the same vanilla Wazuh (zero decoder config)** - the same records arrive as first-class, searchable fields (`event.name`, `user.name`, `harbor.audit.outcome`, `error.type`, `severity.text`):

> OTel index: parsed-field columns across all event types.
04-otel-parsed-fields

Cross-checked on Splunk as well - via OTLP -> collector -> Splunk, every field is auto-extracted with zero parsing config (no `rex`/`spath`), because the OTel record is pure JSON with no syslog header:

> Splunk: OTel events with all fields auto-extracted, no parse step.

10-splunk-otel-autoparsed

Filtering `harbor.audit.outcome: failure` returns exactly the failures with `error.type` populated - impossible on the blob:

> filtering outcome=failure returns exactly the failures.
05-otel-filter-failure

The `config.update` `{from,to}` diff is preserved in `harbor.audit.details`:

>config.update document showing the from/to diff.
06-otel-config-update-fromto

**Satellite side (`component: satellite`)** - register (`flow: ztr`) + hot-reload `config.update` (`source: hot_reload`), over both OTel and syslog/network:

> OTel index filtered to `component: satellite`.
07-otel-satellite-component

> Motivation for shipping both: syslog reaches any SIEM and local/file trails; OTel makes the data usable with no per-format decoder. Detection is out of scope here , with the fields searchable it would be a single alerting monitor, but none is built in this PR.

## Known limitations

- **Satellite audit config is not delivered by GC** - the state artifact omits `app_config.audit`, so a fresh satellite starts with audit off until set in its local config (pre-existing).
- **`daemon` + systemd-journald** - journald does not parse the RFC 5424 APP-NAME/priority; `rsyslog` and the `network` target do.
- **`satellite.deregister.success`** - not reproducible locally (handler deletes the Harbor state repo before emitting -> 404 when no state was pushed). Emit path is wired.
- **`network` is fire-and-forget** (no retry); use `file` where events must not be lost.
- **OTel = visibility, not detection** - the OTel -> indexer path bypasses the rule engine. Brute-force alerting would be one OpenSearch monitor (trivial because the fields exist) but is not built here; GC also already locks accounts after repeated failures.


Aloui-Ikram:
Thanks for the reviews. Summary of what changed since `3ffd496`:

**Commit `2bdc3b8`**
- Sanitize errors before emitting to the audit log on the satellite ZTR path so the token cannot leak through wrapped HTTP error messages (`internal/state/registration_process.go`, plus a unit test for the redaction).
- Validate the `AUDIT_LOG_MAX_*` env values on GC: empty path or negative numbers now cause `log.Fatalf` at startup instead of being passed through to lumberjack.
- Make `X-Forwarded-For` / `X-Real-IP` opt-in via a new env var `AUDIT_TRUST_FORWARDED_HEADERS` (default `false`). The default behaviour is to use the TCP `RemoteAddr` only, so clients cannot spoof `source_ip` on a plain deployment. Documented in the guide.
- Alphabetize the audit keys in `ground-control/.env.example`.

**Commit `517dd7f`**
- Address the follow-up cubic finding on `parseIntEnv` swallowing parse errors: added `parseRequiredIntEnv` which `log.Fatalf`s on a set-but-invalid integer (e.g. `AUDIT_LOG_MAX_SIZE_MB=abc`). The three audit call sites use the new helper; other callers of `parseIntEnv` are unchanged.

**Not addressed (deliberately):** the nit about omitting empty `actor` / `source_ip` fields. The schema in `docs/guides/audit-logging.md` documents these as always-present, and the unit tests assert on their presence, changing this would break the contract for SIEM consumers built against the published schema.

@coderabbitai review
@cubic-dev-ai re-review

Aloui-Ikram:
@bupd @Vad1mo ,thank you for the feedback at the standup last week. Based on your input and the [Harbor audit-logging gist](https://gist.github.com/bupd/ad818658387ed426665300c5524856c7), I reworked the AuditEvent schema into 17 fields.

Before I start writing code, I would like to get your opinion on the shape. You both know Harbor's conventions better than I do, so your feedback would help a lot here.

 If the design works for you, the plan is:

1. Land the schema in this PR
2. Add syslog transport next
3. OTel + CloudEvents envelope after that

### The 17 fields and where each one comes from

| Field | Notes |
|---|---|
| `event_id` | unique ID per event — kept, UUID v4 |
| `timestamp` | event time — kept, UTC RFC3339 |
| `severity` | lets SIEM prioritize without parsing `event_type`; maps directly to syslog PRI and OTel later — OWASP / syslog / OTel |
| `actor` | who triggered the event — kept |
| `actor_type` | distinguish `user` / `robot` / `satellite` / `anonymous` — OWASP via the gist |
| `component` | which side emitted the event (`satellite` / `ground-control`) — was implicit in the filename before |
| `satellite_id` | identifies which satellite the event came from |
| `source_ip` | useful for investigations (brute-force, geo-anomaly). Would prefer to keep it since OWASP and Harbor both do, but open to dropping |
| `user_agent` | helps identify the client (curl vs browser vs satellite binary) — OWASP via the gist |
| `request_id` | correlate multiple events from one HTTP request — OWASP via the gist |
| `operation` | verb: `login`, `create`, `delete`, `update`, `register`, `deregister`, `password_change`, `auth` — Harbor gist, keeps us consistent |
| `resource_type` | noun: `user`, `satellite`, `config`, `session`, `robot` — Harbor gist |
| `resource` | target instance ID — Harbor gist |
| `outcome` | `success` / `failure` as a real field, not buried inside `event_type` — OWASP + Harbor + OTel |
| `reason` | short sanitized failure code. Strawman enum: `invalid_credentials \| token_expired \| token_invalid \| forbidden \| not_found \| rate_limited` — open to suggestions |
| `event_type` | derived as `{resource_type}.{operation}.{outcome}` → e.g. `user.delete.success`. Breaking change vs the current `user.created` naming — flagging it explicitly |
| `details` | free-form map. For `config.changed`: `{ "from": {...}, "to": {...} }` |

### Mapping each field to OTel / syslog / CloudEvents

Using registered semantic conventions where they exist, custom attributes where they do not. No internal field renames will be needed when transports are added later.

| Field | OpenTelemetry | syslog (RFC 5424) | CloudEvents (v1.0.2) |
|---|---|---|---|
| `event_id` | custom attribute (`harbor.audit.event_id`) | STRUCTURED-DATA | `id` (required core attribute) |
| `timestamp` | `Timestamp` field on LogRecord | TIMESTAMP header | `time` (RFC 3339) |
| `severity` | `SeverityText` + `SeverityNumber` fields on LogRecord | PRI severity (`critical`→2, `error`→3, `warning`→4, `info`→6) | `severitytext` + `severitynumber` (Severity [extension](https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/severity.md)) |
| `actor` | `user.id` *(Development stability)* | STRUCTURED-DATA | `authid` (Auth Context [extension](https://github.com/cloudevents/spec/blob/main/cloudevents/extensions/authcontext.md)) |
| `actor_type` | custom attribute (`harbor.audit.actor_type`) | STRUCTURED-DATA | `authtype` (Auth Context extension; enum includes `user`, `service_account`, `system`, `unauthenticated`) |
| `component` | `service.name` + `service.namespace="harbor"` *(Stable)* | APP-NAME header (≤48 ASCII chars) | part of `source` URI |
| `satellite_id` | `service.instance.id` *(Stable)* | STRUCTURED-DATA (custom SD-ID, `name@<PEN>` form per §7) | part of `source` URI |
| `source_ip` | `client.address` *(Stable)* — pair with `network.peer.address` when behind a proxy | STRUCTURED-DATA | `data` |
| `user_agent` | `user_agent.original` *(Stable)* | STRUCTURED-DATA | `data` |
| `request_id` | `http.request.header.x-request-id` *(Stable template attribute, value is `string[]`)* | STRUCTURED-DATA | `data` (or `traceparent` extension only if value is W3C TraceContext format) |
| `operation` | custom attribute (`harbor.audit.operation`) | STRUCTURED-DATA | `data` |
| `resource_type` | custom attribute (`harbor.audit.resource_type`) | STRUCTURED-DATA | `data` |
| `resource` | custom attribute (`harbor.audit.resource`) | STRUCTURED-DATA | `subject` (optional core attribute, fits exactly) |
| `outcome` | custom attribute (`harbor.audit.outcome`) | STRUCTURED-DATA | `data` |
| `reason` | `error.type` *(Stable; must be low-cardinality enum, not free-form text)* | STRUCTURED-DATA | `data` |
| `event_type` | `EventName` field on LogRecord | shortened MSGID (≤32 ASCII chars, e.g. `audit.user.delete`); full string in STRUCTURED-DATA | `type` (prefixed `io.goharbor.satellite.audit.`) |
| `details` | LogRecord Attributes (mind cardinality) | STRUCTURED-DATA (≥480 octets guaranteed, 2 KB recommended) | `data` |

Quick note: the CloudEvents Severity and Auth Context extensions are listed in the spec repo but are not part of the v1.0.2 core. i can  use them, or put those fields in `data` instead ;

---

Could you let me know if this shape works, or if there are fields you would add, drop, or rename? 

Thank you for your time.

---

Update (2026-06-21): a correction to the syslog (RFC 5424) column above. What I actually shipped carries the full canonical JSON in the MSG and leaves STRUCTURED-DATA empty. I did not map each field into STRUCTURED-DATA as the table proposed.

Reason: STRUCTURED-DATA does not actually solve the problem. A vanilla SIEM still needs an operator-side decoder to turn it into searchable fields, so it is no more searchable than JSON-in-MSG. It is also flat `key="value"` only, which does not fit the nested `details` object. JSON-in-MSG keeps the record lossless and parseable in one piece.

Vad1mo:
it is not clear if all fields are meant to be mandatory?

Aloui-Ikram:
oh i  should have made that clearer , sorry.

  Out of the 17, only 8 are always there: `event_id`, `timestamp`,`severity`, `component`, `event_type`, `operation`,
 `resource_type`, `outcome`. The code knows all of these at emit time so they show up on every event.

  The other 9 are optional and omitted when we don't have a value for them. `actor` and `actor_type` drop off on unauthenticated events, `source_ip` / `user_agent` / `request_id` only show up on HTTP-originated events, `satellite_id` only when the event relates to a specific satellite, `resource` only when there's a concrete target, `reason` only on failures, and `details` only when there's extra structured context to attach.

The plan is to omit optional fields when unknown rather than emit empty strings, because OTel, CloudEvents and syslog all do it that way , keeps the internal shape aligned with the transports later on.

bupd:
Review findings:

1. **High: audit logging can be enabled while silently dropping every event if the file cannot be opened.**

   This applies to both `internal/logger/audit.go` and `ground-control/internal/logger/audit.go`. `NewAuditLogger` marks the logger enabled as soon as `Enabled=true` and `FilePath` is non-empty, but `lumberjack` opens the file lazily on the first write. If the directory does not exist or the process lacks write permission, `zerolog.Event.Send()` does not surface that write error to callers, so the service keeps running with `audit.Enabled() == true` while no audit file is produced.

   Realistic case: an operator sets `AUDIT_LOG_PATH=/var/log/ground-control/audit.log` or `file_path=/var/log/harbor-satellite/audit.log`, but the container image/user has no writable `/var/log/...` directory. The PR advertises audit logging as enabled, but auth/user/config events are lost with no startup failure.

   Please fail fast when audit logging is enabled and the destination is not writable, or return an error from logger construction and wire that into startup. A small startup probe that creates the parent directory if that is desired, opens the file with append/create permissions, and closes it would catch the common failure modes before accepting traffic.

2. **Medium: Satellite treats `max_backups=0` and `max_age_days=0` differently from Ground Control and the docs.**

   In `pkg/config/validate.go`, `validateAndEnforceAuditConfig` defaults `MaxBackups` and `MaxAgeDays` when they are `<= 0`. `lumberjack` uses `0` as a meaningful value for these fields (`0` means no limit for backup count/age), and the Ground Control validation now explicitly allows `MAX_BACKUPS >= 0` and `MAX_AGE_DAYS >= 0`.

   Scenario: a Satellite operator configures `max_backups: 0` or `max_age_days: 0` to retain all rotated audit logs, but validation rewrites those to `7` and `30`. That can unexpectedly delete audit history and makes Satellite behavior diverge from GC for the same audit contract.

   Please only default `MaxBackups` and `MaxAgeDays` when they are negative, while keeping `MaxSizeMB <= 0` invalid/defaulted. The test at `pkg/config/validate_test.go` should cover the intentional zero case separately from negative values.

3. **Medium: Satellite audit config changes are not hot-reloaded.**

   `cmd/main.go` constructs the audit logger once from `cm.GetAuditConfig()` and stores it in the request context/local `audit` variable before starting the watcher. Later, hot reload calls `cm.ReloadConfig()` and emits `config.changed` using that original logger, but it never recreates the logger when `app_config.audit` changes.

   Scenario: a Satellite is running with audit disabled and the config is hot-reloaded to enable audit logging. The config manager picks up the new audit settings, but the active logger remains the old no-op logger until process restart, and the enablement change itself is not recorded. The inverse is also possible: disabling audit or changing `file_path` keeps writing through the old logger.

   If `app_config.audit` is intended to be runtime-configurable like the rest of the config, the reload path should rebuild/swap the audit logger when audit fields change. If a restart is required, the guide should say that explicitly and the config change event should not imply live reconfiguration.

4. **Minor: the `source_ip` documentation still says X-Forwarded-For is always honored.**

   `docs/guides/audit-logging.md` says `source_ip` “honors X-Forwarded-For first hop” in the field table. After the spoofing fix, that is only true when `AUDIT_TRUST_FORWARDED_HEADERS=true`; the default is `RemoteAddr`. The later configuration section is correct, but the table should be updated to avoid giving operators the wrong expectation.

5. **Nit: direct audit dependencies are still marked indirect.**

   Both modules import `github.com/google/uuid`, `github.com/rs/zerolog`, and `gopkg.in/natefinch/lumberjack.v2` directly from the new logger packages, but some of those entries remain in indirect require blocks/comments. This does not break the build, but running `go mod tidy` and keeping direct imports in direct require blocks would make the dependency intent clearer.

Verification run locally from the PR head worktree:

```sh
go test ./... -count=1
go test ./... -count=1  # from ground-control/
```

bupd:
Follow-up review, focused on the audit event model and transport compatibility:

1. **High: the current audit event schema is still not syslog / OpenTelemetry ready as a canonical model.**

   The implementation still emits the minimal shape from `internal/logger/audit.go` / `ground-control/internal/logger/audit.go`: `event_id`, `timestamp`, `event_type`, `actor`, `source_ip`, and optional `details`. That is enough for a local JSON log, but it leaves several fields that syslog and OTel need as first-class values either missing or only inferable by parsing strings:

   - `severity` is absent, so syslog PRI severity and OTel `SeverityText` / `SeverityNumber` would need a transport-specific lookup table over `event_type`.
   - `component` is absent from the event, so the record itself does not say whether it came from Satellite or Ground Control. A shipper can infer this from file path today, but OTel/syslog transports should not depend on deployment-specific filenames.
   - `operation`, `resource_type`, `resource`, and `outcome` are absent. The current `event_type` values are not consistently shaped as `{resource}.{operation}.{outcome}` (`user.created`, `config.changed`, `satellite.registered`, `user.password_changed`), so transports cannot reliably derive low-cardinality OTel attributes or syslog structured data without special cases.
   - `reason` is still nested in free-form `details` instead of being a top-level low-cardinality failure code. For OTel this maps much more naturally to `error.type`; burying it in `details` makes alerting and aggregation weaker.
   - HTTP context such as `request_id` and `user_agent` is not captured, which limits incident correlation and makes later CloudEvents/OTel wrapping less useful.

   If this PR is meant to establish the long-term audit contract, I would not land the minimal schema and then retrofit the model later. Please promote the transport-relevant fields into the core audit event struct now, even if some are optional. A good minimum compatible shape would include: `event_id`, `timestamp`, `severity`, `component`, `event_type`, `operation`, `resource_type`, `outcome`, optional `actor`, `actor_type`, `source_ip`, `user_agent`, `request_id`, `resource`, `reason`, `satellite_id`, and `details`.

   With that shape, syslog can map `severity` to PRI, `component` to APP-NAME, `event_type` to MSGID/structured data, and the rest to structured data. OTel can map timestamp/severity directly, use `service.name` / `service.namespace` for component, and attach operation/resource/outcome/reason as stable low-cardinality attributes without parsing `event_type`.

2. **Medium: the docs currently document the minimal JSON shape, so consumers would build against a contract we already know is incomplete.**

   `docs/guides/audit-logging.md` still presents only the minimal JSON example and field table. If the intended direction is the richer syslog/OTel-compatible model discussed in the PR thread, the guide should be updated in the same PR so SIEM/log-pipeline consumers do not lock onto the old shape.

The implementation follow-ups from the previous review look substantially improved: startup now probes audit log writability, Satellite audit settings are detected during hot reload, zero retention values are preserved, `source_ip` docs were corrected, and direct dependencies were cleaned up.

Verification run locally at PR head `09c7b879beb38d1476fc8ed76d0a3f62da05fd1d`:

```sh
go test ./... -count=1
(cd ground-control && go test ./... -count=1)
```

Aloui-Ikram:
@bupd Summary of each commit since your last review (CI green, and E2E-tested on a live GC + satellite stack):

1. `2ba7bce` - reworked `AuditEvent` to the agreed 17-field model. `event_type` is derived as `{resource_type}.{operation}.{outcome}`; `severity` and `component` are derived too; all emit sites, tests and docs updated.
2. `8897eca` - reduced `Log()` cyclomatic complexity below the Codacy threshold (table-driven optional fields, no behaviour change).
3. `3600f0c` - review fixes: actor on the user-authenticated register handler is now the admin, not the satellite; `config.changed` reports `failure` when a hot-reload reconfigure fails instead of always success; satellite-side registration failures now emit `satellite.register.failure`, keeping `satellite.auth.failure` for genuine auth rejections so brute-force alerting stays clean.
4. `da4de4e` - per-request `request_id`: a correlation ID generated per request (reuses a valid inbound `X-Request-ID`), shared by every event from that request.
5. `6a724df` - validate the inbound `X-Request-ID` (safe charset, max length) before it enters the audit log, so a client cannot inject arbitrary strings.
6. `55417a2` - `config.update` now logs `details.changed`: a field-level `from -> to` map with secret values redacted, so a change is auditable field-by-field. `config.create` logs `details.to`, `config.delete` logs `details.from`.

(Also `0a042ad`: merged `main` to resolve a go.mod conflict introduced by #480.)

Could you review the schema before I add the syslog transport? Thanks.

bupd (approved):
/lgtm

PR #450, opened 2026-05-17
fix: add startup jitter to scheduler to prevent thundering herd on fleet restart

### LINKED ISSUE #444: bug: every satellite fires its first sync at T=0 with no jitter, spiking Ground Control on fleet restarts
## The scenario

Imagine a factory floor running 300 satellites. The facility has a power event at 3am. When power returns, all 300 satellites boot together, their containers come up within seconds of each other, and every single one of them immediately fires a registration request and a state fetch at Ground Control - simultaneously.

This is not a hypothesis. It is the guaranteed behavior today because `scheduler.go` calls `launchProcess` at T=0 with no initial delay:

```go
// scheduler.go:60-61
// Run once immediately
s.launchProcess(ctx)
```

Every satellite runs this exact line at startup. They all wake up at the same time, they all call the same endpoint, and Ground Control absorbs all of it at once.

---

## What the load actually looks like

**Today (no jitter), 300 satellites with a 10s sync interval:**

```
Requests to Ground Control
│
600 ┤▐██▌                       ▐██▌                       ▐██▌
    │▐██▌                       ▐██▌                       ▐██▌
    │▐██▌                       ▐██▌                       ▐██▌
    │▐██▌                       ▐██▌                       ▐██▌
  0 ┼──────────────────────────────────────────────────────── time
    T=0 (all satellites boot)   T=10s                      T=20s
```

600 simultaneous requests (300 ZTR registrations + 300 state fetches) every 10 seconds.

**With startup jitter, spreading first fire uniformly across the interval:**

```
Requests to Ground Control
│
 60 ┤▁▂▃▄▅▆▇█▇▆▅▄▃▂▁           ▁▂▃▄▅▆▇█▇▆▅▄▃▂▁           ▁▂▃▄▅▆▇█
    │                                                              
  0 ┼──────────────────────────────────────────────────────── time
    T=0 (satellites boot)       T=10s                      T=20s
```

60 requests/second steady, no spike. Ground Control never sees a wall of traffic.

---

## Root cause in the code

The scheduler's `run()` method fires immediately on start, before the first ticker tick:

```go
// internal/scheduler/scheduler.go:51-81
func (s *Scheduler) run(ctx context.Context) {
    defer s.wg.Done()
    defer s.ticker.Stop()

    s.log.Info()...

    // Run once immediately  ← this is the problem
    s.launchProcess(ctx)

    for {
        select {
        case <-ctx.Done():
            return
        case <-s.ticker.C:
            s.launchProcess(ctx)
        }
    }
}
```

The intent is valid: don't make operators wait a full interval before the satellite does anything. But the side effect is a perfectly synchronized fleet restart storm.

This affects both scheduled processes:
- **`FetchAndReplicateStateProcess`** - hits Ground Control's state endpoint
- **`ZtrProcess`** — hits Ground Control's registration endpoint

Both fire at T=0. In a fleet of N satellites, that is 2N simultaneous requests at startup, then N requests every interval forever.

---

## Why this matters at the edge

Edge deployments have properties that make this worse than in a datacenter:

- **Power events are common**: edge sites often share power with industrial equipment. Momentary power loss brings down a whole facility at once, then everything restarts together.
- **Ground Control is shared**: one Ground Control instance serves many satellites. A spike from 300 satellites is qualitatively different from traffic from 300 independent web users.
- **The rate limiter doesn't help here**: the existing rate limiter in Ground Control (middleware/ratelimit.go) is per-IP, not fleet-wide. All 300 satellites have different IPs, so the rate limiter sees 300 individual clients each making one request - all allowed.
- **The DB pool is finite**: each simultaneous request to Ground Control hits the database. A burst of 600 concurrent DB queries against a connection pool of 25 is 24x oversubscription. Queries queue, timeouts cascade, satellites retry - making the spike worse.

---

## Proposed fix

Add a random initial sleep in `[0, interval)` before the first `launchProcess` call. The ticker-based subsequent fires remain on the fixed interval, so after the first cycle everything is nicely staggered:

```go
func (s *Scheduler) run(ctx context.Context) {
    defer s.wg.Done()
    defer s.ticker.Stop()

    s.log.Info()...

    // Spread satellite startups across the first interval window to avoid
    // a thundering herd when many satellites restart simultaneously.
    jitter := time.Duration(rand.Int63n(int64(s.interval)))
    select {
    case <-ctx.Done():
        return
    case <-time.After(jitter):
    }

    s.launchProcess(ctx)

    for {
        select {
        case <-ctx.Done():
            return
        case <-s.ticker.C:
            s.launchProcess(ctx)
        }
    }
}
```

The jitter is random per satellite process, so a fleet of 300 satellites with a 10s interval will spread their first requests uniformly across [0, 10s] - averaging 30 requests/second instead of 300 at once.

This can optionally be made configurable (`jitter_enabled: true/false` in config) if there are cases where the operator genuinely wants immediate first-fire behavior - though the default should be jittered.

---

## Related

- Ground Control rate limiter: #287 (per-IP, doesn't help with fleet bursts)
- Configurable sync scheduling: #387 (sync windows - complementary, not the same problem)
- Parallelize push job: #312

The rate limiter and sync windows are separate concerns. This issue is specifically about the deterministic synchronized startup that makes every fleet restart a load spike.

bhuvan-somisetty:
Fixes #444.

## What's broken

Every scheduler fires its first execution at T=0 immediately on start, before the first ticker tick:

```go
// scheduler.go current code
s.launchProcess(ctx)   // fires at T=0, no delay
```

This is fine for a single satellite. In a fleet of 300 satellites, a power event at 3am brings every device down at once. When power returns, all 300 boot within seconds of each other. Every scheduler fires at T=0. Ground Control absorbs 600 simultaneous requests (300 registrations + 300 state fetches) in a single burst.

The per-IP rate limiter in Ground Control does not help each satellite has a different IP, so all 300 appear as individual low-frequency clients.

## The spike

```
Requests to Ground Control
│
600 ┤▐██▌                       ▐██▌                       ▐██▌
    │▐██▌                       ▐██▌                       ▐██▌
    │▐██▌                       ▐██▌                       ▐██▌
  0 ┼──────────────────────────────────────────────────────── time →
    T=0 (all satellites boot)   T=10s                      T=20s

600 simultaneous requests saturate a 25-connection DB pool 24x over.
Queries queue, timeouts cascade, satellites retry making the spike worse.
```

## Fix

Draw a random startup delay in `[0, interval)` before the first `launchProcess`. The ticker-based subsequent fires remain on the fixed interval, so after the first cycle every satellite is naturally staggered.

```go
jitter := time.Duration(s.jitterFn(int64(s.interval)))
select {
case <-ctx.Done():
    return
case <-time.After(jitter):
}
// drain any tick that accumulated during the jitter window
select {
case <-s.ticker.C:
default:
}
s.launchProcess(ctx)
```

## After the fix

```
Requests to Ground Control
│
 60 ┤▁▂▃▄▅▆▇█▇▆▅▄▃▂▁           ▁▂▃▄▅▆▇█▇▆▅▄▃▂▁
    │
  0 ┼──────────────────────────────────────────── time →
    T=0 (satellites boot)       T=10s

~60 req/s steady. Ground Control never sees a wall.
```

## How the jitter works

```mermaid
sequenceDiagram
    participant S1 as Satellite 1
    participant S2 as Satellite 2
    participant S3 as Satellite 3
    participant GC as Ground Control

    Note over S1,S3: Power restored all satellites boot simultaneously

    S1->>S1: jitter = rand(0, 10s) = 2s
    S2->>S2: jitter = rand(0, 10s) = 6s
    S3->>S3: jitter = rand(0, 10s) = 9s

    Note over S1: T=2s  first fire
    S1->>GC: register + fetch state ✓

    Note over S2: T=6s first fire
    S2->>GC: register + fetch state ✓

    Note over S3: T=9s first fire
    S3->>GC: register + fetch state ✓

    Note over S1,S3: After first cycle, all satellites tick independently no re-synchronization
```

## Scheduler lifecycle before vs after

```mermaid
flowchart TD
    A[Scheduler.Start called] --> B{BEFORE fix}
    B --> C[launchProcess at T=0 immediately]
    C --> D[wait interval]
    D --> E[launchProcess]
    E --> D

    A --> F{AFTER fix}
    F --> G[draw jitter in 0 to interval]
    G --> H{ctx cancelled\nduring jitter?}
    H -->|yes| I[exit cleanly ✓]
    H -->|no| J[drain accumulated tick]
    J --> K[launchProcess at T=jitter]
    K --> L[wait interval]
    L --> M[launchProcess]
    M --> L
```

## Implementation detail ticker drain

After sleeping through the jitter window, one tick may have accumulated in the ticker channel (if `jitter > 0` and the ticker fired during the sleep). Without draining it, the second execution would happen almost immediately after the first:

```
Without drain (jitter=9s, interval=10s):
  T=9s   → first execution
  T=10s  → ticker fires (1s later nearly back-to-back)
  T=20s  → third execution

With drain:
  T=9s   → first execution (tick drained)
  T=19s  → second execution (full interval after first)
  T=29s  → third execution
```

## Making jitterFn injectable

`jitterFn` is a struct field (default: `rand.Int63n`) so tests can override it without a global mock or interface change:

```go
// in tests zero jitter for deterministic immediate execution
sched.jitterFn = func(int64) int64 { return 0 }

// in tests verify jitter spread
sched.jitterFn = func(n int64) int64 { return n / 2 }
```

This keeps the production default clean while making scheduler behaviour fully testable.

## Changes

| File | Description |
|------|-------------|
| `internal/scheduler/scheduler.go` | Add `jitterFn` field; apply jitter with ctx-cancel and ticker drain before first execution |
| `internal/scheduler/scheduler_test.go` | Zero jitter in two tests that rely on immediate first fire; add three new jitter tests |

## Tests added

| Test | What it covers |
|------|----------------|
| `TestScheduler_JitterBounds` | `jitterFn` is called with `interval` as the upper bound — jitter stays in `[0, interval)` |
| `TestScheduler_ContextCancelDuringJitter` | Cancelling context during jitter window exits cleanly; process never fires |
| `TestScheduler_JitterSpreadsFirstFire` | 10 schedulers with distinct fixed offsets produce first-fire times spread ≥ half the interval |

## Test plan

- [x] `go test -v -count=1 ./internal/scheduler/...` all 8 tests pass
- [x] `go build ./...` full build clean
- [x] `task lint-report` clean

## Summary by CodeRabbit

* **New Features**
  * Schedulers now include randomized jitter on first execution, distributing startup times across the configured interval rather than clustering all executions simultaneously.

[Review Change Stack](https://app.coderabbit.ai/change-stack/container-registry/harbor-satellite/pull/450?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


internal/scheduler/scheduler.go
```diff
@@ -57,7 +64,37 @@ func (s *Scheduler) run(ctx context.Context) {
...
+	case <-jitterTimer.C:
+	}
+	// Drain any tick that fired during the jitter window so the first
+	// scheduled interval begins cleanly after the initial execution.
+	select {
+	case <-s.ticker.C:
+	default:
+	}
+
 	s.launchProcess(ctx)
```
shri771, line 96: The stagger only applies to this first fire the ticker's phase is fixed at construction time, so a fleet that boots together re-synchronizes on the very next tick. It could be better to keep them spread, reset the ticker here so it's phase re-anchors to now before the first run. That also makes the preceding drain block redundant.

shri771 (changes requested):
Left a suggestion.

shri771:
This is a sizable change to a core path the whole fleet depends on, so it might be worth a maintainer's read on the approach before going deeper in the issue.

internal/scheduler/scheduler.go
bhuvan-somisetty: Good point, fixed — resetting the ticker right after the jitter delay now so the phase re-anchors to that moment instead of staying pinned to construction time. Dropped the drain block too since Reset already takes care of it. Thanks for catching this, tests still pass.

bhuvan-somisetty:
Totally fair — this touches the scheduling loop every satellite runs on, so happy to wait for a maintainer to weigh in on the approach before I go further. Pushed a fix for the ticker re-anchoring issue in the meantime.

bupd:

PR #462, opened 2026-05-26
chore: upgrade Go toolchain to 1.26.2 and drop redundant toolchain directive

### LINKED ISSUE #461: Upgrade to go 1.26.x & Zot to v3
Currently we are using go 1.24 - we should be using go 1.26 

also update zot dependency to v3.

would also be good to update all dependencies - while updating these.

bhuvan-somisetty:
Go 1.26.2 has been out for a while now, and we were still pinned to 1.24 in both modules. This bumps the `go` directive in `go.mod` and `ground-control/go.mod` to 1.26.2, updates all CI workflow steps to match, and removes the now-redundant `toolchain go1.24.3` line from `ground-control/go.mod`  that directive only makes sense when the toolchain differs from the `go` line, which it no longer does.

The `GoVersion` constant in `internal/version/version.go` is updated to reflect the new baseline.

`go mod tidy` was run on both modules after the upgrade; only the `go` directive changed in the lock files, so there are no dependency additions or removals from this PR.

Go 1.26 is a drop-in compatible release. Nothing in either module uses features that changed between 1.24 and 1.26, so existing behavior is unchanged. CI on Linux (where the Zot syscall surface is fully supported) should go green without modification.

Closes #461.

## Summary by CodeRabbit

* **Chores**
  * Upgraded Go toolchain from 1.24 to 1.26.5 across build configurations, Go module requirements, and CI/CD workflows.
  * Updated reported build Go version to 1.26.
  * Pinned GitHub Actions to specific commit SHAs (including checkout/setup-go/task, goreleaser, SBOM tooling, cosign installer, and PR labeling) for deterministic runs.
* **Security**
  * Standardized workflow permissions to `read-all` at the top level and restricted job permissions to `contents: read`.

### DIFF
--- a/.github/actions/publish-and-sign/action.yaml
+++ b/.github/actions/publish-and-sign/action.yaml
@@ -27,7 +27,7 @@ runs:
     - name: Setup Go
       uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
       with:
-        go-version: "1.26.2"
+        go-version: "1.26.5"
 
     - name: Install Task
       uses: arduino/setup-task@b91d5d2c96a56797b48ac1e0e89220bf64044611 # v2.0.0
--- a/.github/actions/publish-and-sign/action.yaml
+++ b/.github/actions/publish-and-sign/action.yaml
@@ -25,12 +25,12 @@ runs:
   using: "composite"
   steps:
     - name: Setup Go
-      uses: actions/setup-go@v5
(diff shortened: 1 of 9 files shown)


bhuvan-somisetty:
Fixed both — Taskfile.yml and the publish-and-sign composite action now use 1.26.2 too. While I was in there I also pinned the remaining unpinned actions (setup-go, setup-task, cosign-installer in the composite action, and actions/labeler which runs on pull_request_target) to SHAs, matching the rest of this PR — figured that's likely what Codacy's security scan was flagging.

bupd:

Issue #453, opened 2026-05-20
bug: Unit test suite fails on Windows hosts and inside minimal container environments


gcharpe1604:
## Description
The unit test suite crashes or fails to pass depending on the host OS or testing environment:
1. **Windows Host:** `internal/secure/config_test.go` asserts a strict POSIX file mode check (`0o600`), which fails on Windows due to different file attribute mapping.
2. **Containerized Linux Host:** `internal/identity/device_linux.go` expects `/etc/machine-id` to exist and read successfully. In minimal Docker/Alpine runners without this file present, it propagates a hard filesystem read error and crashes the device tests instead of reporting the component as unavailable.

## Steps to Reproduce
1. **Windows:** Run `go test ./internal/secure/...` on a Windows machine.
2. **Alpine Container:** Run `docker run --rm -v "${PWD}:/app" -w /app golang:1.24-alpine go test ./...` in a container lacking `/etc/machine-id`.

## Expected Behavior
All tests should pass or skip gracefully depending on host capabilities, without failing the entire test build.

## Actual Behavior
1. `TestConfigEncryptor_EncryptToFile` fails on Windows due to permission mode mismatch.
2. `TestLinuxDeviceIdentity_GetMachineID` and `TestLinuxDeviceIdentity_GetBootID` fail with filesystem error `no such file or directory`.

## Environment
- OS: Windows 11 / Linux (Alpine)
- Tool version: Go 1.24

## Additional Context
Standard practice for hardware identifier checks (like network MAC addresses or CPU serials) is to return `ErrComponentUnavailable` and skip the tests rather than failing the build.

PR #471, opened 2026-05-29
fix: replace fmt.Print with proper error wrapping and structured logging

### LINKED ISSUE #470: bug: fmt.Print/Printf in internal packages bypasses zerolog, silently discards error details
## Summary

Several internal packages use bare `fmt.Print`/`fmt.Printf` calls instead of the project's zerolog structured logger. This has two concrete problems:

1. **Errors are silently discarded** — `internal/state/fetcher.go:228` prints only the string `"Error in unmarshalling"` without including `err`, so the actual unmarshal failure reason is lost entirely.
2. **Bypasses structured logging** — these calls ignore the `--json-logging` flag and log-level filtering, so they always emit to stdout as plain text regardless of operator configuration.

## Affected locations

| File | Line | Issue |
|------|------|-------|
| `internal/state/fetcher.go` | 228 | `fmt.Print("Error in unmarshalling")` — `err` is available but not logged |
| `internal/state/state_process.go` | 578 | `fmt.Printf("Error in getting repository and image name: %v", err)` |
| `ground-control/internal/server/config_handlers.go` | 255, 269 | `fmt.Println("Could not list configs: ", err)` inside HTTP handlers |

`cmd/main.go` and `ground-control/main.go` usages are expected (logger not yet initialized) and are **not** part of this issue.

## Impact

- Operators running with `--json-logging` receive mixed output (JSON lines interleaved with plain `fmt` output), breaking log aggregation pipelines.
- When `fetcher.go` fails to unmarshal state, the actual JSON parse error is never surfaced — the operator sees only `Error in unmarshalling` with no actionable detail.
- Log-level filtering (e.g. `warn`-only) does not suppress these lines.

## Expected fix

Replace each call with the appropriate zerolog call:

```go
// internal/state/fetcher.go:228 — include the error
log.Error().Err(err).Msg("error unmarshalling state")

// internal/state/state_process.go:578
log.Error().Err(err).Msg("error getting repository and image name from artifact")

// ground-control/internal/server/config_handlers.go:255,269
log.Error().Err(err).Msg("could not list configs")
```

Each package already imports zerolog and has a package-level logger variable, so this is a straightforward fix with no new dependencies.

## Steps to reproduce

1. Run satellite with `--json-logging` and trigger an unmarshal failure (e.g. corrupt state artifact).
2. Observe that stdout contains a plain `Error in unmarshalling` line rather than a JSON log entry, and that the actual parse error is absent.

bhuvan-somisetty:
Closes #470

---

When I was digging through the state processing code while working on the upgrade branch, I spotted something that could silently make debugging a nightmare in production: three places where bare `fmt.Print`/`fmt.Println` calls were used instead of the structured zerolog logger the rest of the codebase uses. One of them was particularly bad the actual error was available but never included in the output.

## What was wrong

```
internal/state/fetcher.go:228        → fmt.Print("Error in unmarshalling")
internal/state/state_process.go:578  → fmt.Printf("Error in getting repository and image name: %v", err)
ground-control/internal/server/config_handlers.go:255,269 → fmt.Println("Could not list configs/get config: ", err)
```

The satellite ships with a `--json-logging` flag so operators can pipe logs into their aggregation stack (Loki, Fluentd, etc.). These `fmt` calls write raw text directly to stdout and completely bypass that. If an unmarshal fails mid-sync, the operator sees a plain string with no timestamp, no log level, no context fields and in the `fetcher.go` case, not even the actual error message.

### Before: what the logging pipeline looked like

```mermaid
flowchart TD
    A[State sync triggered] --> B[fetchGroupState]
    B --> C[extractArtifactJSON]
    C --> D{artifacts.json found?}
    D -->|yes| E[json.Unmarshal]
    E -->|error| F["fmt.Print('Error in unmarshalling')"]
    F --> G["return nil, err"]
    G --> H[Caller logs the error ✓]
    F -.->|bypasses| Z["zerolog pipeline ✗\n(no JSON, no level, no fields)"]

    style F fill:#ff6b6b,color:#fff
    style Z fill:#ff6b6b,color:#fff
```

### After: clean error propagation

```mermaid
flowchart TD
    A[State sync triggered] --> B[fetchGroupState]
    B --> C[extractArtifactJSON]
    C --> D{artifacts.json found?}
    D -->|yes| E[json.Unmarshal]
    E -->|error| F["return nil, fmt.Errorf('unmarshal state: %w', err)"]
    F --> G[Caller receives wrapped error with full context]
    G --> H["log.Error().Err(err).Msg(...) ✓"]

    style F fill:#51cf66,color:#fff
    style H fill:#51cf66,color:#fff
```

### Logging pipeline: before vs after

```mermaid
sequenceDiagram
    participant Satellite
    participant fmt as fmt (stdout)
    participant zerolog as zerolog (structured)
    participant Loki as Log aggregator

    Note over Satellite,Loki: BEFORE broken path
    Satellite->>fmt: fmt.Print("Error in unmarshalling")
    fmt-->>Loki: ❌ raw text, no level, no error details
    Note over fmt,Loki: Aggregator can't parse this

    Note over Satellite,Loki: AFTER fixed path
    Satellite->>zerolog: return fmt.Errorf("unmarshal state: %w", err)
    zerolog->>Loki: {"level":"error","error":"unmarshal state: unexpected end of JSON","message":"..."}
    Note over zerolog,Loki: ✓ structured, queryable, complete
```

## Changes

**`internal/state/fetcher.go`** - `FromJSON` was printing `"Error in unmarshalling"` without including `err`. Replaced with `fmt.Errorf("unmarshal state: %w", err)` so the actual parse failure propagates up to the caller that already has a logger.

**`internal/state/state_process.go`** - `ProcessState` was calling `fmt.Printf` before returning the error. Removed the printf and wrapped the error with the artifact repository name as context, which is far more useful when debugging a bad artifact.

**`ground-control/internal/server/config_handlers.go`**  Two handlers (`listConfigsHandler`, `getConfigHandler`) were using `fmt.Println` while every other handler in that same file uses the stdlib `log` package. Brought them in line with the rest of the file.

No behavior change errors still surface and handlers still respond correctly. The fixes only affect how (and whether) failure context reaches the log pipeline.

## Testing

Verified the ground-control server package passes `go vet` cleanly. The satellite build currently fails on Windows due to a pre-existing Zot/syscall issue (#453), unrelated to these changes. Both modified files pass `gofmt -e` without errors.

---

**Bhuvan Somisetty**

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logging and diagnostic detail for configuration lookup/listing failures.
  * Added contextual error wrapping for state unmarshal failures and repository/image name extraction.
* **CI/CD**
  * Updated lint and test workflows to use a dedicated golangci-lint cache and added a unit-test job with coverage uploads.
  * Disabled Go setup caching across several workflow jobs.
* **Documentation / Website**
  * Added comprehensive contributor guide and new blog content plus navigation and styling updates (including a new blog layout).
* **Chores**
  * Refreshed e2e Harbor Docker Compose configuration for updated images and startup behavior.


shri771 (changes requested):
@bhuvan-somisetty,  Heads up: #448  is introducing a zerolog-backed logger into this same file for the other three handlers, so the file is moving toward zerolog overall. It would be cleaner to use that here too, instead of stdlib log, so all the handlers use the same logger once #448 is merged.

bhuvan-somisetty (commented):
Thanks @shri771  good catch. Switched both handlers to use `zerolog.Ctx(r.Context())` so they're consistent with what #448 is wiring in. Also promoted zerolog to a direct dependency in `ground-control/go.mod` via `go mod tidy`. The `getConfigHandler` line now also carries the config name as a structured field for better debuggability.

shri771:
@bhuvan-somisetty, I rechecked #448 and it doesn't actually attach a zerolog logger to the request context, not zerolog.Ctx). So zerolog.Ctx(r.Context()) here will hit the disabled default logger and the errors will silently disappear. Probably simplest to use log.Println.

Aloui-Ikram:
Hey @bhuvan-somisetty  , since this is about #448   @shri771 's right.
 In ground-control nothing puts a zerolog logger into the request context (the middleware only adds the user), so zerolog.Ctx(r.Context()) would just hit the disabled default and the errors would silently disappear.
One small clarification so #448 isn't confusing: the zerolog logger I added there is the audit logger a separate one that writes security events to the audit file. For normal errors in this file, #448 just uses log.Println like the other handlers.
So log.Println here is the right call , it actually keeps us consistent.
   Thanks

bhuvan-somisetty:
Good catch, you're right — switched both back to `log.Print`/`log.Printf` so the errors actually get logged instead of vanishing into the disabled default logger. Will move these to zerolog properly once #448 lands and wires the logger into the context.

bupd:

PR #481, opened 2026-06-04
fix(ci): provide envsubst for cosign installer

bupd:
## Summary
- add an envsubst fallback before running sigstore/cosign-installer
- keep the publish-and-sign action working on self-hosted runners without gettext installed

## Verification
- go test ./...
- go test ./... (ground-control)
- git diff --check
- verified the fallback expands `/var/home/bupd/.cosign`

## Summary by CodeRabbit

* **Chores**
  * Improved reliability of the publish-and-sign workflow step by adding support for environments where the `envsubst` utility is not pre-installed.


bupd:
Closing in favor of #482, which contains the same fix with the required DCO signoff.

PR #493, opened 2026-06-25
fix (#475): delete Harbor artifact before database transaction in deleteGroupHandler

### LINKED ISSUE #475: bug: deleteGroupHandler commits DB transaction before deleting Harbor artifact, leaving orphaned OCI state artifacts on failure
## Summary

`deleteGroupHandler` in `ground-control/internal/server/group_handlers.go` commits the database transaction **before** attempting to delete the group's OCI state artifact from Harbor. If the artifact deletion fails, the group is permanently gone from the database but its state artifact lives in Harbor indefinitely.

## The bug

In `group_handlers.go` the ordering is:

```go
// line 307 — DB transaction committed, group deleted from DB
if err := tx.Commit(); err != nil { ... }
committed = true

// line 318 — artifact deletion attempted AFTER the commit
err = utils.DeleteArtifact(utils.ConstructHarborDeleteURL(groupName, "group"))
if err != nil {
    HandleAppError(w, err) // Returns error to caller...
    return                 // ...but DB is already committed!
}
```

If `utils.DeleteArtifact` returns an error:
- the group no longer exists in the database
- the group's OCI state artifact is still present in Harbor
- the client receives an HTTP error response, making it appear the deletion failed entirely
- a subsequent retry hits `GetGroupByName → 404` and can never clean up the Harbor artifact
- the orphaned artifact persists until someone manually deletes it from Harbor

## Contrast with the correct pattern

`deleteConfigHandler` in `config_handlers.go` does this correctly: it deletes the Harbor artifact **first**, and only then removes the record from the database:

```go
// line 403 — artifact deleted first
if err := utils.DeleteArtifact(utils.ConstructHarborDeleteURL(configName, "config")); err != nil {
    return // DB untouched if artifact deletion fails
}

// line 412 — DB deletion only happens if artifact deletion succeeded
if err := q.DeleteConfig(r.Context(), configObject.ID); err != nil { ... }
```

## Impact

- Orphaned group state artifacts accumulate in Harbor's `satellite` project, wasting storage and causing confusion when listing artifacts.
- If a new group with the same name is later created, it may encounter the stale artifact in Harbor.
- Any satellites that polled the group state artifact between commit and failed artifact deletion see inconsistent state: the DB reports the group gone, but the OCI artifact is still pullable.
- The caller receives an error and may retry, but each retry 404s at `GetGroupByName` and can never resolve the orphan.

## Related code

| File | Line | Operation |
|------|------|-----------|
| `group_handlers.go` | 307 | `tx.Commit()` - DB group deletion committed |
| `group_handlers.go` | 318 | `utils.DeleteArtifact(...)` - happens **after** commit |
| `config_handlers.go` | 403 | `utils.DeleteArtifact(...)` - happens **before** DB delete (correct) |

## Fix

Move `utils.DeleteArtifact(...)` to before `tx.Commit()` (or at minimum before `committed = true`), mirroring the approach in `deleteConfigHandler`. Since `DeleteArtifact` calls Harbor's REST API and cannot be rolled back, the safest ordering is:

1. Attempt Harbor artifact deletion if it fails, return error without touching DB
2. Delete the group from the database
3. Commit the transaction

This matches the config handler's defensive ordering and ensures the caller always gets a truthful response about what actually changed.

Prat260104:
Fixes #475

## Summary

`deleteGroupHandler` was calling `utils.DeleteArtifact` after `tx.Commit`. If the Harbor API call failed (network timeout, Harbor downtime), the group was permanently removed from the database but its OCI artifact remained in Harbor with no way to clean it up, since the group no longer existed in the database to retry against.

## Changes

Moved the Harbor artifact deletion to before the transaction begins, following the same pattern already used in `deleteConfigHandler`:

1. Verify group exists
2. Delete Harbor artifact — if this fails, database is unchanged
3. Begin transaction, update satellite states, delete group from DB, commit

If Harbor deletion succeeds but the database transaction fails, the client receives an error and can retry. Harbor artifact deletion is idempotent (404 is acceptable on retry).

## Testing

- All existing unit tests pass
- golangci-lint clean (0 issues)
- Build successful

## Notes

PostgreSQL's `ON DELETE CASCADE` on `satellite_groups.group_id` handles junction table cleanup automatically when the group is deleted, so no changes to the satellite association logic were needed.

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/493?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved group deletion handling to reduce the chance of partial cleanup when removing a group.
  * Updated error messaging during deletion to be clearer and more consistent.

### DIFF
--- a/ground-control/internal/server/group_handlers.go
+++ b/ground-control/internal/server/group_handlers.go
@@ -144,6 +144,31 @@ func (s *Server) deleteGroupHandler(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	groupName := vars["group"]
 
+	q := s.dbQueries
+
+	group, err := q.GetGroupByName(r.Context(), groupName)
+	if err != nil {
+		log.Println(err)
+		err := &AppError{
+			Message: "Error: Group Not Found",
+			Code:    http.StatusNotFound,
+		}
+		HandleAppError(w, err)
+		return
+	}
+
+	// Delete Harbor artifact before modifying database
+	err = utils.DeleteArtifact(utils.ConstructHarborDeleteURL(groupName, "group"))
+	if err != nil {
+		log.Println(err)
+		err := &AppError{
+			Message: "Error: Failed to delete group state artifact",
+			Code:    http.StatusInternalServerError,
+		}
+		HandleAppError(w, err)
+		return
+	}
+
 	tx, err := s.db.BeginTx(r.Context(), nil)
 	if err != nil {
 		log.Printf("Error starting transaction: %v", err)
@@ -155,7 +180,7 @@ func (s *Server) deleteGroupHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	q := s.dbQueries.WithTx(tx)
+	q = s.dbQueries.WithTx(tx)
 
 	committed := false
 	defer func() {
@@ -171,17 +196,6 @@ func (s *Server) deleteGroupHandler(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 	}()
-
-	group, err := q.GetGroupByName(r.Context(), groupName)
-	if err != nil {
-		log.Println(err)
-		err := &AppError{
-			Message: "Error: Group Not Found",
-			Code:    http.StatusNotFound,
-		}
-		HandleAppError(w, err)
-		return
-	}
 
 	// Remove the group from all associated satellites
 	satellites, err := q.GroupSatelliteList(r.Context(), group.ID)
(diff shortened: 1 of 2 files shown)


Prat260104:
Hi @Vad1mo @bupd , 

this PR fixes issue #475 — **deleteGroupHandler** was calling **utils.DeleteArtifact** after **tx.Commit**, which could leave orphaned **Harbor artifacts** if the deletion failed post-commit. Moved it to before the transaction begins, following the same pattern already used in **deleteConfigHandler**.
Would love your feedback whenever you have time!

bupd:
invalid and has conflicts

PR #513, opened 2026-07-04
fix: validate satellite and config names in setSatelliteConfig handler

AdeshDeshmukh:
## Description

The `setSatelliteConfig` handler accepted `req.Satellite` and `req.ConfigName` without validating them via `utils.IsValidName()`. Every sibling handler already performs this check. This inconsistency allowed invalid names (uppercase, empty, special characters) to reach the database layer and downstream OCI artifact operations.

### Changes

1. **`internal/groundcontrol/server/config_handlers.go`** — Added two `utils.IsValidName()` guard blocks after `DecodeRequestBody` and before `BeginTx`. Invalid names are rejected with a clear 400 Bad Request before any database transaction is opened.

2. **`internal/groundcontrol/server/config_handlers_test.go`** — Added three test cases covering malformed body, uppercase satellite name, and empty config name. All are pure unit tests — no DB mocking needed since validation fires before any database call.

### Verification

- `task lint-report`: 0 issues
- Full test suite: 582 tests passed, 0 failures, 0 regressions
- `go vet`: No issues

## Additional context

**Sibling handlers with same validation:** `registerSatelliteHandler`, `createConfigHandler`, `updateConfigHandler`, `addSatelliteToGroup`, `removeSatelliteFromGroup`.

**Validator:** `IsValidName()` at `internal/groundcontrol/utils/helper.go:313` — 1-255 chars, lowercase + digits + ._-, must start with letter/digit.

### DIFF
--- a/internal/groundcontrol/server/config_handlers.go
+++ b/internal/groundcontrol/server/config_handlers.go
@@ -461,6 +461,22 @@ func (s *Server) setSatelliteConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if !utils.IsValidName(req.Satellite) {
+		HandleAppError(w, &AppError{
+			Message: fmt.Sprintf(invalidNameMessage, "satellite"),
+			Code:    http.StatusBadRequest,
+		})
+		return
+	}
+
+	if !utils.IsValidName(req.ConfigName) {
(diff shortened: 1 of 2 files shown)


bupd:
invalid without discussion

PR #520, opened 2026-07-10
fix(state): terminate collectResults loop using labeled break

### LINKED ISSUE #519: bug: bare break in collectResults only exits select, not the for loop
### Summary

In `internal/state/state_process.go`, the `collectResults` function uses a bare `break` statement inside a loop containing a `select` block. In Go, a `break` statement without a label only terminates the innermost `select`, `switch`, or `for` statement. As a result, the loop never terminates normally, causing the goroutine to block indefinitely even after all results are successfully processed.

---

### Code Reference
[`internal/state/state_process.go#L320-L322`](file:///home/cotishq/lfx/harbor-satellite/internal/state/state_process.go#L320-L322)

cotishq:
- Fixes: #519 

## Description

This PR fixes a bug in `collectResults` where a bare `break` statement was used inside a loop containing a `select` block. In Go, a bare `break` only terminates the innermost `select` statement, rather than the enclosing `for` loop.

### What changed:
- Added a loop label `outer` to the `for` loop.
- Updated the break condition to use `break outer` to terminate the collection loop once all state and config fetcher results have been processed.
This allows the function to exit cleanly, execute the error-handling block, and return a clean success/failure status rather than blocking indefinitely until context cancellation.

## Additional context

- Confirmed all unit tests pass locally: `go test ./...`

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/520?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/satellite/state/state_process.go
+++ b/internal/satellite/state/state_process.go
@@ -281,6 +281,7 @@ func (f *FetchAndReplicateStateProcess) collectResults(
 	receivedStateFetchers := 0
 	receivedConfigFetcher := false
 
+outer:
 	for {
 		select {
 		case <-ctx.Done():
@@ -315,7 +316,7 @@ func (f *FetchAndReplicateStateProcess) collectResults(
 		}
 
 		if receivedStateFetchers == expectedCount && receivedConfigFetcher {
-			break
+			break outer
 		}
 	}
 


cotishq:
Closing this PR as it doesn't fix a real bug.

PR #547, opened 2026-07-29
feat(scheduler): track per-run history and rename IsComplete to Shoul…

### LINKED ISSUE #528: `IsComplete` in a `Process`` shows the _scheduling process_ status rather than the state for the _instance of the process_
Currently, the `IsComplete()` function defined for the interface `Process`,
```go
// Process represents a process that can be scheduled
type Process interface {
	// Name returns the name of the process
	Name() string

	// Execute runs the process
	Execute(ctx context.Context) error

	// IsRunning returns true if the process is running
	IsRunning() bool

	// ShouldStop returns true if the process scheduling should be stopped
	IsComplete() bool
}
```

Shows us whether the scheduling of this process should stop, instead I propose that it should show the status of that instance of a process, for example, if the `reporting_process` schedules 5 times, then the `isComplete()` function should show the _completeness_ of each of the _instance_ of the process.

Such that, we could further enhance this feature and show,
- Process Logs per run
- Failure reports for each run
- and more

Something like,

```go
type RefreshCredentialProcess struct {
	name       string
	isRunning  bool
	isComplete bool
	cm         *config.ConfigManager
	log        *zerolog.Logger

	mu sync.RWMutex
}

// In this process object, we could have a complete() such that
func (s *RefreshCredentialProcess) complete() {
	s.mu.Lock()
	defer s.mu.Unlock()
	s.isComplete = true
}

// Which could be used like
func (s *RefreshCredentialProcess) Execute(ctx context.Context) error {
	s.start()
	defer s.complete()

}
```

While this is not a big issue for now, but this moves forwards in making a Process run itself agnostic from the scheduler, and allowing us to aggregate and show more performance/qol data as required.

bhavyaKhatri2703:
- Fixes: #528  

## Description
1. Renamed `IsComplete()` → `ShouldStop()`
Makes explicit that this signals the scheduler should stop scheduling the process entirely (unchanged behavior for ZtrProcess, FetchAndReplicateStateProcess, SpiffeZtrProcess, StatusReportingProcess).

2. Added per-run history tracking on Scheduler
similar in spirit to isComplete, but per-instance, not per-process-lifecycle: each `Execute() `call now records a `RunResult{StartedAt, FinishedAt, Err}` in a bounded ring buffer (last 10 runs), exposed via `History()` and `LastRun()`

3. Every run now logs one line (Info on success, Warn on failure) with duration/success/error fields

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/547?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added scheduler run history, including start and finish times, duration, success status, and errors.
  * Added access to the latest run and retained run history.
  * Limited retained history to the latest 10 runs.

* **Improvements**
  * Updated process stopping behavior across scheduler processes for clearer lifecycle handling.
  * Scheduler logging now reflects whether runs succeed or fail.

### DIFF
--- a/internal/satellite/scheduler/process.go
+++ b/internal/satellite/scheduler/process.go
@@ -16,5 +16,5 @@ type Process interface {
 	IsRunning() bool
 
 	// ShouldStop returns true if the process scheduling should be stopped
-	IsComplete() bool
+	ShouldStop() bool
 }
--- a/internal/satellite/scheduler/scheduler.go
+++ b/internal/satellite/scheduler/scheduler.go
@@ -8,8 +8,26 @@ import (
 	"sync"
 	"time"
 
+	"github.com/container-registry/harbor-satellite/pkg/config"
 	"github.com/rs/zerolog"
 )
+
+// RunResult records the outcome of a single execution of a scheduled process.
+type RunResult struct {
+	StartedAt  time.Time
+	FinishedAt time.Time
+	Err        error
+}
+
+// Duration returns how long the run took.
+func (r RunResult) Duration() time.Duration {
+	return r.FinishedAt.Sub(r.StartedAt)
+}
+
+// Success reports whether the run completed without error.
+func (r RunResult) Success() bool {
+	return r.Err == nil
+}
 
 // Scheduler manages the execution of processes with configurable intervals.
 type Scheduler struct {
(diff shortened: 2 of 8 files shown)


bupd:
invalid without discussion

PR #551, opened 2026-07-29
fix: enforce strict @every format in cron validator

### LINKED ISSUE #472: bug: cron schedule validator rejects all valid @every intervals, silently overriding user config
## Summary

The satellite's config validator and its scheduler speak completely different interval formats, and neither component tells the user. Any custom sync interval a user sets in `@every` format is silently discarded and replaced with the 30-second default. Any standard cron expression that passes validation crashes the scheduler at startup.

## What's happening

There are two separate components involved:

**Validator** (`pkg/config/validate.go:78`) uses `cron.ParseStandard` from `robfig/cron/v3`:

```go
func isValidCronExpression(cronExpression string) bool {
    if _, err := cron.ParseStandard(cronExpression); err != nil {
        return false
    }
    return true
}
```

`cron.ParseStandard` only accepts the standard 5-field cron format: `* * * * *`. It does NOT recognize `@every`, `@daily`, or any descriptor-style expressions.

**Scheduler** (`internal/scheduler/scheduler.go:161`) uses its own parser that ONLY accepts `@every <duration>`:

```go
func parseEveryExpr(expr string) (time.Duration, error) {
    const prefix = "@every "
    if !strings.HasPrefix(expr, prefix) {
        return 0, fmt.Errorf("unsupported format: must start with %q", prefix)
    }
    return time.ParseDuration(strings.TrimPrefix(expr, prefix))
}
```

**The defaults** (`pkg/config/constants.go`) use `@every` format:

```go
const DefaultFetchAndReplicateCronExpr string = "@every 00h00m30s"
const DefaultZTRCronExpr               string = "@every 00h00m05s"
const DefaultHeartbeatCronExpr         string = "@every 00h00m30s"
```

## The three broken scenarios

```
┌─────────────────────────────────────────────────────────────────────┐
│ User config value       │ Validator result  │ Scheduler result       │
├─────────────────────────┼───────────────────┼────────────────────────┤
│ "@every 5m"             │ INVALID (rejected)│ Never reached — config │
│ (natural format, based  │ Silently replaced │ replaced with 30s      │
│ on the defaults)        │ with "@every 30s" │ default before startup │
├─────────────────────────┼───────────────────┼────────────────────────┤
│ "*/5 * * * *"           │ VALID (accepted)  │ FAILS at runtime with  │
│ (standard cron, as used │                   │ "unsupported format:   │
│ in the test suite)      │                   │ must start with @every"│
├─────────────────────────┼───────────────────┼────────────────────────┤
│ "" (not set)            │ Replaced with     │ Works — default was    │
│                         │ "@every 00h00m30s"│ never validated        │
└─────────────────────────┴───────────────────┴────────────────────────┘
```

The only scenario that actually works end-to-end is leaving the interval empty, because the default value is applied *after* validation and never goes through `isValidCronExpression`.

## Impact

- Operators who read the defaults (`@every 00h00m30s`) and model their config after them (e.g., `@every 5m` for a 5-minute interval) will have their setting silently reset to 30 seconds. This can cause unexpected load on Ground Control: a fleet configured for hourly syncs actually syncs every 30 seconds.
- Operators who try to use standard cron syntax (as the test case in `validate_test.go` suggests is valid) will get a runtime startup failure in the scheduler.
- The warning emitted for a "bad" `@every` interval is `"invalid schedule provided for state_replication_interval, using default schedule @every 00h00m30s"` — which looks like the operator made a typo, not like a validator bug.

## Steps to reproduce

```json
{
  "app_config": {
    "ground_control_url": "http://127.0.0.1:8080",
    "state_replication_interval": "@every 5m"
  }
}
```

Start the satellite with the above config. Observe:
1. Warning: `"invalid schedule provided for state_replication_interval, using default schedule @every 00h00m30s"`
2. Satellite syncs every 30 seconds instead of every 5 minutes.

## Root cause

`isValidCronExpression` was written for standard cron but the scheduler was built around `@every <duration>`. The two formats were never reconciled. The test suite masked this because `"0 * * * *"` passes validation but was never run through the scheduler in a test.

## Suggested fix

Replace `cron.ParseStandard` in `isValidCronExpression` with a parser that accepts `@every <duration>` — the only format the scheduler actually handles. The simplest approach:

```go
func isValidScheduleExpression(expr string) bool {
    const prefix = "@every "
    if !strings.HasPrefix(expr, prefix) {
        return false
    }
    _, err := time.ParseDuration(strings.TrimPrefix(expr, prefix))
    return err == nil
}
```

Alternatively, align the scheduler to also accept standard cron expressions if that broader format is desired long-term.

**Bhuvan Somisetty**

Akarshkushwaha:
- Fixes: #472

## Description
This PR addresses a bug where the configuration validator silently accepted invalid standard cron expressions (which crash the scheduler) while seemingly overriding valid `@every <duration>` schedules. 

The issue was caused by using `cron.ParseStandard` from `robfig/cron/v3`, which is too permissive for the internal scheduler that strictly expects the `@every <duration>` format.

This replaces the `robfig/cron/v3` parser with a strict `time.ParseDuration` check to ensure the config validator only accepts the `@every ` format supported by the internal scheduler.

## Additional context
- I removed the now-unused `robfig/cron/v3` parser logic from `isValidCronExpression`.
- I have verified that `task test` passes with this new strict validation logic.
- I confirmed via manual scratch testing that standard cron strings like `*/5 * * * *` are now successfully rejected and correctly fallback to the default.

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/551?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Schedule settings now support duration-based expressions using the `@every` format.

* **Bug Fixes**
  * Improved validation for replication and satellite registration intervals.
  * Invalid schedule values continue to fall back to safe defaults with warnings.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -438,7 +438,6 @@ require (
 	github.com/prometheus/procfs v0.20.1 // indirect
 	github.com/rcrowley/go-metrics v0.0.0-20250401214520-65e299d6c5c9 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
-	github.com/robfig/cron/v3 v3.0.1
 	github.com/rubenv/sql-migrate v1.8.1 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/sagikazarmark/locafero v0.12.0 // indirect
--- a/pkg/config/validate.go
+++ b/pkg/config/validate.go
@@ -7,9 +7,9 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/container-registry/harbor-satellite/internal/satellite/registry"
-	"github.com/robfig/cron/v3"
 	"github.com/rs/zerolog"
 )
 
@@ -203,13 +203,14 @@ func enforceAuditRotation(f *SyslogAuditFile) []string {
 	return warnings
 }
 
-// isValidCronExpression checks the validity of a cron expression.
-func isValidCronExpression(cronExpression string) bool {
-	if _, err := cron.ParseStandard(cronExpression); err != nil {
+// isValidScheduleExpression checks the validity of a schedule expression.
+func isValidScheduleExpression(expr string) bool {
+	const prefix = "@every "
+	if !strings.HasPrefix(expr, prefix) {
 		return false
 	}
-
-	return true
+	_, err := time.ParseDuration(strings.TrimPrefix(expr, prefix))
+	return err == nil
 }
 
 // validateAndEnforceLogLevel validates log level and defaults to info if invalid.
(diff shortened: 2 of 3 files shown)


bupd:

PR #553, opened 2026-07-31
fix(harbor): refactor GetClient to return errors instead of panicking

### LINKED ISSUE #552: Refactor Ground Control's Harbor API client initialization to return errors instead of panicking
Description:
Is your feature request related to a problem? Please describe. Currently, when internal/groundcontrol/harbor/client.go initializes the Harbor client (GetClientByConfig and GetClient), it uses panic(err) if harbor.NewClientSet returns an error (for example, if the Harbor URL is invalid or malformed). Panicking inside a server component like Ground Control is a Go anti-pattern because it will crash the entire process upon misconfiguration, rather than gracefully propagating the error.

Describe the solution you'd like Refactor GetClientByConfig and GetClient to return (*v2client.HarborAPI, error) instead of panicking. All downstream callers (like those in robot.go, project.go, replication.go) should be updated to receive the error and propagate it up the stack using standard Go error handling.

Describe alternatives you've considered None. Crashing the server process on startup/client initialization due to a URL config is inherently unsafe.

Additional context I would be happy to submit a Pull Request to fix this issue!

shellyco-code:
Description:
Fixes #552 

What this PR does / why we need it: When internal/groundcontrol/harbor/client.go initializes the Harbor client (GetClientByConfig and GetClient), it previously used panic(err) if harbor.NewClientSet returned an error. This is a Go anti-pattern for server applications like Ground Control, as a simple client misconfiguration or invalid URL would cause the entire process to crash.

This PR refactors GetClientByConfig and GetClient to return (*v2client.HarborAPI, error) instead. All downstream callers (in robot.go, project.go, and replication.go) have been properly updated to capture the error and safely propagate it up the stack using standard error handling.

Which issue(s) this PR fixes: Fixes #(Replace with Issue Number)

Special notes for your reviewer: I used a module-level clientErr error to safely capture and return the error resulting from clientOnce.Do(). All callers were updated to check if err != nil and return the error gracefully.

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/553?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/groundcontrol/harbor/client.go
+++ b/internal/groundcontrol/harbor/client.go
@@ -10,11 +10,12 @@ import (
 
 var (
 	client     *v2client.HarborAPI
+	clientErr  error
 	clientOnce sync.Once
 )
 
 // Returns Harbor v2 client
-func GetClient() *v2client.HarborAPI {
+func GetClient() (*v2client.HarborAPI, error) {
 	clientOnce.Do(func() {
 		cfg := env.GC.Harbor
 		clientConfig := &harbor.ClientSetConfig{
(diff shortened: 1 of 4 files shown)


bupd:
would be good to have a discussion before creating this - Thanks

Issue #542, opened 2026-07-27
LFX Mentorship Term 3: Air-Gapped Peer-to-Peer OCI Image Distribution


bupd:
# LFX Mentorship Term 3 proposal

## Summary

Enable trusted Harbor Satellite instances on the same isolated network to share OCI images and artifacts directly. When a Satellite needs an artifact that is absent from its local registry, it should be able to obtain it from an eligible peer without Internet, Harbor, or Ground Control connectivity.

## Problem

Air-gapped deployments can bootstrap images from offline OCI layouts or packages, but this must be repeated for each Satellite. If one trusted Satellite already has a needed artifact, another Satellite cannot currently reuse that local copy. This increases the operational overhead, storage-media handling, and time needed to recover or expand an isolated edge site.

## Goals

- Add an opt-in peer-to-peer distribution mode for trusted Satellites on an isolated network.
- Support explicitly configured peer endpoints in the initial implementation, so the feature works without Ground Control or external discovery.
- Locate an eligible peer for a requested OCI artifact and copy its manifest and referenced blobs into the requesting Satellite's local registry.
- Preserve OCI digest verification and avoid publishing incomplete artifacts.
- Provide bounded retries, timeouts, concurrency controls, structured logs, and metrics for peer attempts and transferred bytes.
- Validate the feature with automated multi-Satellite, air-gapped integration tests and operator documentation.

## Proposed work

1. Produce an ADR/design covering the peer model, artifact lookup flow, trust boundary, integrity guarantees, retries, and headless-mode behavior.
2. Add validated configuration for feature enablement, trusted peer endpoints, timeouts, retry/backoff, and concurrency limits.
3. Implement peer eligibility and artifact-availability checks using OCI registry APIs.
4. Implement digest-verified peer-to-peer artifact transfer into the embedded local registry.
5. Add observability for successful transfers, misses, peer failures, and bytes transferred.
6. Add unit tests, an end-to-end multi-Satellite test environment, and user/operator documentation.

## Acceptance criteria

- With Internet and Ground Control unavailable, Satellite B can obtain a requested OCI image or artifact from configured Satellite A.
- The imported manifest and every referenced blob are verified by digest before the artifact is available locally.
- A missing or unreachable peer fails predictably, records useful diagnostics, and does not corrupt local registry state.
- The feature is disabled by default and existing replication/bootstrap flows remain unchanged.
[...]

PR #554, opened 2026-07-31
feat: add P2P configuration and validation for air-gapped peer replication

Norway-02:
*Related to #542*

### Summary

While looking into how Satellites handle missing artifacts in air-gapped deployments, I noticed a gap in the current replication flow. If a Satellite needs an image but has no Ground Control or internet access, it has no way to retrieve it - - even if another Satellite on the same local network already has the artifact.

This PR doesn't implement peer-to-peer transfers yet. Instead, it introduces the configuration model and architectural foundation needed to support Satellite-to-Satellite artifact sharing in follow-up changes.

### The problem

```text
Satellite A (has image)          Satellite B (needs image)
        |                                 |
        |                          [no Ground Control]
        |                          [no internet]
        |                                 |
        X -------- currently impossible --------> X
```

### What changed

**Configuration**

Added a new `P2PConfig` section to `AppConfig` along with validation.

The validation currently checks:
- Peer URLs are valid.
- Acquisition timeout is between 5 seconds and 5 minutes.
- Maximum concurrent transfers is between 1 and 20.

If `max_concurrent_transfers` isn't configured, it defaults to 5 with a warning, following the same pattern used by `validateRegistryFallbackConfig`.

Since Go's standard JSON unmarshalling doesn't parse duration strings like `"30s"` into `time.Duration`, I also added a custom `UnmarshalJSON` implementation for the configuration.

**Replication**

Introduced `PeerAwareReplicator`, which wraps the existing `Replicator` interface.

At this stage it simply delegates to the existing implementation, but it establishes a clean extension point for peer-aware routing without changing `BasicReplicator` or the current replication pipeline.

### Why split it this way?

I wanted to get the configuration model and abstraction boundary reviewed before introducing networking code.

The planned follow-up work builds on this foundation:
1. Peer discovery and availability checks.
2. OCI artifact transfer.
3. Fallback handling when peers cannot satisfy a request.

Keeping these pieces separate should make each review much smaller and easier to reason about.

### Backward compatibility

This change does not modify the existing replication behavior.

When `p2p.enabled` is false (the default), the replication path is unchanged.

### Files changed

- `pkg/config/config.go` - - `P2PConfig` and custom `UnmarshalJSON`
- `pkg/config/validate.go` - -  `validateP2PConfig`
- `pkg/config/getters.go` - - thread-safe getters (`GetP2PPeers` returns a copy)
- `pkg/config/modifiers.go` - - `SetP2PConfig`
- `internal/satellite/state/peer_replicator.go` - - `PeerAwareReplicator`
- `pkg/config/config_test.go` - - configuration and JSON parsing tests
- `internal/satellite/state/peer_replicator_test.go` - -  delegation and constructor tests

### Testing

Verified with:

`go build ./...`  
`go test ./...`  
`go test -race ./...`  

All completed successfully.

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/internal/satellite/state/peer_replicator.go
+++ b/internal/satellite/state/peer_replicator.go
@@ -1,0 +1,53 @@
+package state
+
+import (
+	"context"
+
+	"github.com/container-registry/harbor-satellite/pkg/config"
+)
+
+// PeerAwareReplicator wraps an existing Replicator and attempts peer acquisition before fallback.
+// In this initial architectural milestone, it strictly delegates all entities to the fallback replicator.
+type PeerAwareReplicator struct {
+	p2pCfg   config.P2PConfig
+	fallback Replicator
+
+	localURL      string
+	localUsername string
+	localPassword string
+	useUnsecure   bool
+}
+
+// NewPeerAwareReplicator creates a new PeerAwareReplicator that delegates to the provided fallback.
+// Panics if the fallback replicator is nil.
+func NewPeerAwareReplicator(
+	cfg config.P2PConfig,
+	fallback Replicator,
+	localURL string,
+	localUsername string,
+	localPassword string,
+	useUnsecure bool,
+) *PeerAwareReplicator {
+	if fallback == nil {
+		panic("fallback replicator cannot be nil")
+	}
+
+	return &PeerAwareReplicator{
+		p2pCfg:        cfg,
+		fallback:      fallback,
+		localURL:      localURL,
+		localUsername: localUsername,
+		localPassword: localPassword,
+		useUnsecure:   useUnsecure,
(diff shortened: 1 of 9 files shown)


bupd:
invalid - created without discussion

PR #561, opened 2026-08-01
fix: publish satellite state artifact when a satellite has no group states

### LINKED ISSUE #537: Publish cleared satellite state artifacts when no groups remain
## Summary

Update the Ground Control state-artifact publication flow so satellites receive a published artifact even when their group-state list is empty.

Currently, `createOrUpdateSatStateArtifact` in `internal/groundcontrol/server/helpers.go` returns without publishing when `states` is empty. This can leave a newly registered satellite with no state artifact and can leave a satellite that was removed from its final group consuming a stale, previously published group list.

## Required changes

- Update the satellite state-artifact/handler flow to publish `SatelliteStateArtifact` with an explicit empty `states` list and the applicable configuration state when there are no group states.
- Verify the registration flow for a satellite with zero groups and the removal flow for a satellite’s final group both publish the cleared artifact.

## Affected areas

- `internal/groundcontrol/server/helpers.go`
- Ground Control satellite registration and group-removal handlers/call sites

## Acceptance criteria

- Registering a satellite with no groups publishes a valid state artifact with `states: []`.
- Removing a satellite from its final group publishes a valid state artifact with `states: []`.
- The artifact retains the expected configuration state.
- Add focused regression coverage for both flows.

## Backlinks

- Originating PR: https://github.com/container-registry/harbor-satellite/pull/536
- Review discussion: https://github.com/container-registry/harbor-satellite/pull/536#discussion_r3629326683
- Requested by: @vg006

krishpinto:
- Fixes: #537

## Description

This PR makes Ground Control publish a satellite state artifact even when the satellite has no group states.

`CreateOrUpdateSatStateArtifact` returned early when `states` was empty, so nothing was published at all. A satellite registered with no groups never received a state artifact, and a satellite removed from its final group kept reading the artifact published before the removal, so it carried on replicating a group it no longer belonged to.

### Changes

- Removed the early return in `internal/groundcontrol/utils/helper.go` so the empty case publishes with an empty states list.
- Dropped `omitempty` from `States` in `internal/groundcontrol/models/models.go`. With the tag in place an empty list is left out of the JSON entirely and the satellite reads that the same as a missing field.
- `getGroupStates` returns a nil slice when a satellite has no groups, and nil marshals to `null` rather than `[]`, so nil is normalized to an empty slice before the artifact is built.
- Added `internal/groundcontrol/utils/helper_test.go`, which is the first test file in that package.

## Additional context

`SatelliteStateArtifact` is marshalled in one place only, and the satellite side unmarshals into a separate type in `internal/satellite/state/state.go`, so dropping `omitempty` does not affect the consumer. The struct carries a `swagger:model` annotation but is not in the committed spec, so nothing needed regenerating.

The tests sit at the unit level. `CreateOrUpdateSatStateArtifact` pushes to a live registry and there is no registry fixture in the repo, so the registration and group removal handlers cannot reach the publish step in a unit test. I checked that both tests fail when the fix is reverted. Happy to add a fixture if you would rather have coverage at the handler level.

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/561?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Satellite state artifacts are now handled consistently when no state data is available.
  * Invalid satellite names are validated before processing.
  * Empty state data can now proceed through artifact creation and validation.

### DIFF
--- a/internal/groundcontrol/server/helpers.go
+++ b/internal/groundcontrol/server/helpers.go
@@ -33,19 +33,36 @@ func isConfigInUse(ctx context.Context, q *database.Queries, config database.Con
 	return len(satellites) > 0, nil
 }
 
-func createOrUpdateSatStateArtifact(ctx context.Context, satelliteName string, states []string, configName string) error {
-	if satelliteName == "" {
-		return fmt.Errorf("the satellite name must be at least one character long")
-	}
+// satelliteStateArtifact is the published form of SatelliteStateArtifact. The
+// generated model tags states as omitempty, so an empty list is left out of the
+// JSON entirely and a satellite cannot tell a cleared state from an unchanged
+// one. The published artifact needs to carry an explicit "states": [].
+type satelliteStateArtifact struct {
+	Config string   `json:"config,omitempty"`
+	States []string `json:"states"`
+}
+
+// buildSatelliteStateArtifact assembles the artifact published for a satellite.
+// A satellite with no group states still needs one, otherwise it either never
+// receives an artifact or keeps consuming a previously published group list, so
+// a nil slice is normalized to an empty one rather than marshaling to null.
+func buildSatelliteStateArtifact(states []string, configName string) satelliteStateArtifact {
 	if states == nil {
 		states = []string{}
 	}
 
-	artifact := SatelliteStateArtifact{
+	return satelliteStateArtifact{
 		States: states,
 		Config: utils.AssembleConfigState(configName),
 	}
-	data, err := json.Marshal(artifact)
+}
+
+func createOrUpdateSatStateArtifact(ctx context.Context, satelliteName string, states []string, configName string) error {
+	if satelliteName == "" {
+		return fmt.Errorf("the satellite name must be at least one character long")
+	}
+
+	data, err := json.Marshal(buildSatelliteStateArtifact(states, configName))
 	if err != nil {
 		return fmt.Errorf("failed to marshal satellite state artifact to JSON: %w", err)
 	}
--- a/internal/groundcontrol/server/helpers.go
+++ b/internal/groundcontrol/server/helpers.go
@@ -34,11 +34,11 @@ func isConfigInUse(ctx context.Context, q *database.Queries, config database.Con
 }
 
 func createOrUpdateSatStateArtifact(ctx context.Context, satelliteName string, states []string, configName string) error {
-	if len(states) == 0 {
-		return nil
-	}
 	if satelliteName == "" {
 		return fmt.Errorf("the satellite name must be at least one character long")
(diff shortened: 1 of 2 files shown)


internal/groundcontrol/utils/helper_test.go
```diff
@@ -0,0 +1,71 @@
...
+			states []string
+		}{
+			{name: "nil states", states: nil},
+			{name: "empty states", states: []string{}},
+		}
+
+		for _, tt := range tests {
+			t.Run(tt.name, func(t *testing.T) {
+				err := CreateOrUpdateSatStateArtifact(context.Background(), "sat-1", tt.states, "cfg")
+				require.ErrorContains(t, err, "HARBOR_URL")
```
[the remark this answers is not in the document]
krishpinto: My bad G, The old subtest only proved the early return was gone and never executed the normalization. I pulled the artifact construction into buildSatelliteStateArtifact so it can be asserted without a registry, and it now covers nil, empty and populated states. Renamed the other subtest to say what it actually guards.

bupd (changes requested):
please resolve conflicts 

thanks for the contribution

internal/groundcontrol/models/models.go
```diff
@@ -6,7 +6,7 @@ import "github.com/container-registry/harbor-satellite/pkg/config"
 //
 // swagger:model SatelliteStateArtifact
 type SatelliteStateArtifact struct {
-	States []string `json:"states,omitempty"`
+	States []string `json:"states"`
```
bupd: this line should not be changed

internal/groundcontrol/utils/helper.go
bupd, line 1: changing here would be right thing to do

bupd (changes requested):

krishpinto:
Hey bupd — rebased and reworked this. Turns out the OAS refactor that landed right around when you reviewed removed `models.go` and moved `CreateOrUpdateSatStateArtifact` out of `utils/helper.go` entirely, so the fix now lives in `server/helpers.go`'s `createOrUpdateSatStateArtifact` instead. Didn't touch the generated model tag — just removed the early return and normalized nil states before building the artifact. Tests and lint pass, ready for another look whenever you get a chance.

PR #564, opened 2026-08-01
perf: replace O(n^2) array scans in updateStateMap with O(1) map lookups

### LINKED ISSUE #558: performance: updateStateMap performs repeated linear lookups when removing stale state entries
## Description

`updateStateMap` checks whether existing state URLs are still present by repeatedly calling the helper function `contains()`.

Because `contains()` performs a linear scan, repeated lookups inside the cleanup loop result in quadratic time complexity.

## Current Behavior

```go
func contains(slice []string, item string) bool {
    for _, s := range slice {
        if s == item {
            return true
        }
    }
    return false
}
```

The helper is repeatedly invoked while iterating through existing state entries.

## Impact

For satellites managing a large number of groups or repositories, the cleanup phase performs unnecessary repeated linear searches, increasing CPU time during every synchronization cycle.

## Suggested Fix

Build a `map[string]struct{}` from the incoming URL list before iterating.

This reduces membership checks to O(1) and lowers the overall cleanup complexity from O(n²) to O(n).

## Files Affected

- `internal/satellite/state/state_process.go`

doraem-on:
Fixes #558

## Summary

`updateStateMap` performs repeated membership checks while reconciling the incoming state with the existing state map.

This change replaces repeated linear lookups with hash-based lookups, reducing the cost of membership checks and improving scalability for larger state sets.

## Why

The previous implementation relied on the helper function `contains()`, which performs a linear scan for every lookup.

As the number of tracked states grows, these repeated searches increase the amount of work performed during each synchronization cycle.

Using hash-based lookups allows membership checks to complete in constant time.

## Changes

- Built lookup maps from the incoming and existing state collections.
- Replaced repeated slice searches with map lookups.
- Removed the now-unused `contains()` helper.

## Testing

- No functional behavior was changed.
- Existing test suites continue to pass.
- Verified that state reconciliation produces the same results after the refactor.

## Additional Context

This optimization was identified while reviewing the satellite state synchronization path. It improves scalability without changing the behavior of the reconciliation logic.

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/564?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Performance**
  * Improved state processing efficiency, especially when handling larger sets of state URLs, resulting in faster updates.
* **Bug Fixes**
  * Preserved existing behavior when adding, retaining, or removing state URLs.
* **Reliability**
  * Added coverage for state updates involving URL additions, removals, and unchanged state, helping ensure consistent results.


bupd (approved):
/lgtm

PR #571, opened 2026-08-04
fix(scheduler): add opt-in startup jitter to spread a fleet's first sync

### LINKED ISSUE #444: bug: every satellite fires its first sync at T=0 with no jitter, spiking Ground Control on fleet restarts

1991santhu:
Fixes #444.

## The problem

`Scheduler.run` calls `launchProcess` at T=0 with no delay:

```go
// Run once immediately
s.launchProcess(ctx)
```

Every satellite executes that line at startup. A fleet that boots together — after a power event, a rolling restart, or a coordinated rollout — issues its first registration and state fetch in the same instant, and because the interval is identical everywhere the alignment persists on every subsequent tick rather than dispersing.

## The change

A bounded random delay before the **first** run only. Subsequent runs remain ticker-driven, so steady-state behaviour is untouched.

```go
s.WithStartupJitter(s.GetInterval())
```

**It is opt-in and defaults to zero.** That is deliberate: defaulting it on changes the semantics of every existing caller, and it broke `TestStop_WaitsForInflightProcess` when I tried it — that test encodes run-immediately as expected behaviour, and it is right to. So the library default is unchanged and the fleet-facing call sites opt in.

`internal/satellite/satellite.go` enables it for the three schedulers that talk to Ground Control — registration (ZTR), state replication, and status reporting — each bounded by its own interval, which spreads first execution uniformly across one interval.

The wait honours context cancellation, so shutting down during the delay exits promptly instead of blocking for up to a full interval.

## Testing

`go test ./internal/satellite/...` passes. New cases cover: the default is zero; the first run is immediate without jitter; the first run is delayed with jitter; cancellation during the wait exits promptly and does not execute the process; and a negative bound clamps to zero rather than panicking in `rand.Int63n`.

## Notes for review

- Using each scheduler's interval as the bound means a satellite may wait up to one full interval before its first sync. For the intervals in the issue's scenario that is seconds; if any deployment uses a long registration interval, a smaller fixed cap may be preferable and I am happy to change it.
- `math/rand` rather than `crypto/rand` — this is load spreading, not a security boundary.
- I have not reproduced the 300-satellite scenario from the issue; the reasoning is from the scheduler code. Happy to adjust if the fleet behaviour differs from what I have assumed.

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/571?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added configurable startup jitter to stagger initial scheduler executions and reduce simultaneous activity.
  * Added a configuration option to disable startup jitter when immediate execution is preferred.
  * Startup delays respect cancellation, handle invalid values safely, and preserve regular scheduling intervals.
  * Default behavior remains unchanged when startup jitter is not configured.

* **Tests**
  * Added coverage for default behavior, delayed execution, cancellation, interval preservation, and invalid settings.

### DIFF
--- a/internal/satellite/satellite.go
+++ b/internal/satellite/satellite.go
@@ -65,6 +65,9 @@ func (s *Satellite) Run(ctx context.Context) error {
 			log.Error().Err(err).Msg("Failed to create ZTR scheduler")
 			return err
 		}
+		// Spread the first registration across one interval. Without this every
+		// satellite in a fleet that boots together registers at the same instant.
+		ztrScheduler.WithStartupJitter(ztrScheduler.GetInterval())
 		s.schedulers = append(s.schedulers, ztrScheduler)
 		ztrScheduler.Start(ctx)
 	}
@@ -79,6 +82,8 @@ func (s *Satellite) Run(ctx context.Context) error {
 		log.Error().Err(err).Msg("Failed to create state replication scheduler")
 		return err
 	}
+	// Spread the first state fetch across one interval, for the same reason.
+	stateScheduler.WithStartupJitter(stateScheduler.GetInterval())
 	s.schedulers = append(s.schedulers, stateScheduler)
 	stateScheduler.Start(ctx)
 
@@ -96,6 +101,8 @@ func (s *Satellite) Run(ctx context.Context) error {
 		log.Error().Err(err).Msg("Failed to create status report scheduler")
 		return err
 	}
+	// Spread the first status report across one interval, for the same reason.
+	statusScheduler.WithStartupJitter(statusScheduler.GetInterval())
 	s.schedulers = append(s.schedulers, statusScheduler)
 	statusScheduler.Start(ctx)
 
--- a/internal/satellite/satellite.go
+++ b/internal/satellite/satellite.go
@@ -67,7 +67,9 @@ func (s *Satellite) Run(ctx context.Context) error {
 		}
 		// Spread the first registration across one interval. Without this every
 		// satellite in a fleet that boots together registers at the same instant.
-		ztrScheduler.WithStartupJitter(ztrScheduler.GetInterval())
+		if !s.cm.IsStartupJitterDisabled() {
+			ztrScheduler.WithStartupJitter(ztrScheduler.GetInterval())
+		}
 		s.schedulers = append(s.schedulers, ztrScheduler)
 		ztrScheduler.Start(ctx)
 	}
@@ -83,7 +85,9 @@ func (s *Satellite) Run(ctx context.Context) error {
 		return err
 	}
 	// Spread the first state fetch across one interval, for the same reason.
-	stateScheduler.WithStartupJitter(stateScheduler.GetInterval())
+	if !s.cm.IsStartupJitterDisabled() {
+		stateScheduler.WithStartupJitter(stateScheduler.GetInterval())
+	}
 	s.schedulers = append(s.schedulers, stateScheduler)
 	stateScheduler.Start(ctx)
 
@@ -102,7 +106,9 @@ func (s *Satellite) Run(ctx context.Context) error {
 		return err
 	}
 	// Spread the first status report across one interval, for the same reason.
-	statusScheduler.WithStartupJitter(statusScheduler.GetInterval())
+	if !s.cm.IsStartupJitterDisabled() {
+		statusScheduler.WithStartupJitter(statusScheduler.GetInterval())
+	}
 	s.schedulers = append(s.schedulers, statusScheduler)
 	statusScheduler.Start(ctx)
 
--- a/internal/satellite/scheduler/scheduler.go
+++ b/internal/satellite/scheduler/scheduler.go
@@ -2,9 +2,10 @@ package scheduler
 
 import (
 	"context"
+	"crypto/rand"
 	"errors"
 	"fmt"
-	"math/rand"
+	"math/big"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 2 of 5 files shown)


1991santhu:
Thanks, all four were fair. Fixed three; the fourth was already dealt with in a force-push that this review predates.

Buffered tick: correct, and my bug. The ticker starts at construction and keeps running while the jitter elapses, so a tick could already be waiting when the first run finished and fire a second one straight away. The scheduler resets the ticker after the wait now, so the gap between the first and second run is a full interval. There's a regression test using an interval shorter than the jitter bound, so a buffered tick is guaranteed rather than hoped for.

Unconditional jitter at the call sites: this was the sharpest one. You're right that it contradicted the framing. The library default is off, but `satellite.go` turned it on everywhere with no way out, so a standalone satellite paid up to a full interval before its startup-critical state fetch, for a fleet problem it doesn't have. Added `disable_startup_jitter` to app config and all three call sites honour it.

Flaky assertion: also correct. Asserting no execution within 200ms of a one-hour uniform bound flakes about once in 18,000 runs. Replaced with a tight bound, an assertion that the run isn't synchronous with `Start`, and `Eventually` for completion. Ran it with `-count=3`.

gosec G404: already fixed before this review ran. The call carries `//nolint:gosec` with a reason, matching `internal/groundcontrol/middleware/ratelimit.go`.

Full `./internal/...` suite passes.

PR #644, opened 2026-08-21
fix(state): keep incomplete groups out of crash checkpoints

### LINKED ISSUE #533: Bug: Nil Entities State Save causes full re-replication on crash restart
## What

If the satellite process crashes during a sync cycle where a new group is detected, the state file is persisted with `entities: null` for that new group. On restart, this triggers a full re-replication of all images in that group, wasting bandwidth.

---

## Why

During sync, `updateStateMap` adds new groups to `f.stateMap` with `Entities: nil`. Currently, the state is saved to disk immediately after this (lines 119–123), before the goroutines have a chance to fetch the image lists.

If the process crashes before the goroutines finish and call `SaveState` (line 485), the state file on disk contains `entities: null`. On restart, `LoadState` populates `f.stateMap` with a nil slice. When `GetChanges` is called (line 187), it checks `if oldEntites == nil` and schedules every image in the group for re-replication.

---

## Proposed Approach

Remove the early save at lines 119–123 and move it to after `collectResults` completes:

```go
// Remove lines 119-123 (early save)

// After line 143 (collectResults):
err := f.collectResults(ctx, stateFetcherResults, configFetcherResult, len(f.stateMap), &log)
if f.stateFilePath != "" && changed {
    if saveErr := SaveState(f.stateFilePath, f.stateMap, f.currentConfigDigest); saveErr != nil {
        log.Warn().Err(saveErr).Msg("Failed to persist state after sync")
    }
}
return err
```

This ensures we never persist incomplete state to disk:
- If a crash happens during goroutine execution, the state file retains the old state. On restart, the new group is re-detected and processed from scratch.
- If all groups are deleted, `changed` is true but no goroutines launch. The post-collect save correctly writes an empty state to disk.

---

## Out of Scope

Changing how `GetChanges` handles nil vs empty slices, or modifying `SaveState` JSON serialization.

AnouarMohamed:
- Fixes: #533

## Description

New group entries are added to `stateMap` with `Entities == nil` before reconciliation starts. That nil value means "pending", but it was being serialized to `state.json` like a completed checkpoint. If the satellite stopped in that window, restart loaded an incomplete group and treated it as having no prior entity state.

This keeps pending groups in memory but leaves them out of persisted checkpoints until reconciliation succeeds. Successfully reconciled groups with no artifacts now use an explicit empty slice, so real empty states are still persisted. Removed groups are still flushed immediately, and completed groups keep their incremental crash-recovery checkpoint.

Tests cover:
- mixed complete and pending groups
- completed empty groups
- restart recovery
- stale group replacement

## Additional context

Validated with:
- `go test ./... -count=1`
- `go test -race ./... -count=1`
- `go vet ./...`
- `task build`
- `task lint`
- `task vuln`
- `task --force generate:ground-control` (no diff)
- recovery-focused tests repeated 50 times

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved state persistence and reconciliation reliability by preventing incomplete groups from being saved.
  * Preserved completed groups that contain no entities.
  * Improved recovery behavior so only incomplete groups are requeued.
  * Removed stale group state before replacement completes.
  * Ensured empty entity results are represented consistently.
  * Correctly selects local or remote storage based on configuration.

* **Tests**
  * Added coverage for storage selection, persistence behavior, recovery, and empty-state handling.

### DIFF
--- a/internal/satellite/state/state_process.go
+++ b/internal/satellite/state/state_process.go
@@ -117,13 +117,14 @@ func (f *FetchAndReplicateStateProcess) Execute(ctx context.Context) error {
 
 	f.mu.Lock()
 	changed := f.updateStateMap(satelliteState.States)
-	f.mu.Unlock()
-
+	var persistErr error
 	// Persist state if groups were added, removed, or swapped
 	if f.stateFilePath != "" && changed {
-		if err := f.PersistState(); err != nil {
-			log.Warn().Err(err).Msg("Failed to persist state after group changes")
-		}
+		persistErr = f.persistStateLocked()
+	}
+	f.mu.Unlock()
+	if persistErr != nil {
+		log.Warn().Err(persistErr).Msg("Failed to persist state after group changes")
 	}
 
 	// Create channels for results
--- a/internal/satellite/state/state_process.go
+++ b/internal/satellite/state/state_process.go
@@ -63,8 +63,10 @@ func NewFetchAndReplicateStateProcess(cm *config.ConfigManager, stateFilePath st
 }
 
 type StateMap struct {
-	url      string
-	State    StateReader
+	url   string
+	State StateReader
+	// A nil Entities slice marks a group that has not completed reconciliation.
+	// A completed group with no artifacts uses a non-nil empty slice.
 	Entities []Entity
 }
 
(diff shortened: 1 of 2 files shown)


omlahore (commented):
The recovery test at state_process_test.go:410 still calls NewFetchAndReplicateStateProcess with three args, but #648 added storeRoot, so the state tests don't build on this head and the test workflow never ran here to catch it. The pending group also comes back with nil Entities after a restart, and that test checks it, so GetChanges still replicates it in full like the null entry did. I guess that's cheap anyway, oras.Copy skips images the OCI layout already has.

PR #645, opened 2026-08-21
fix(store): continue replicating remaining artifacts after per-entity failure

### LINKED ISSUE #625: bug: first failing artifact in Replicate permanently blocks all later artifacts in the group
## Description

`BasicReplicator.Replicate` in `internal/satellite/state/replicator.go` iterates a group's entities and returns on the first failure, so every entity after the failing one is skipped:

```go
desc, err := remote.Get(src, pullOpts...)
if err != nil {
    log.Error().Msgf("Failed to fetch image descriptor: %v", err)
    return err
}
```

The same pattern appears at the `remote.Write` call (line 171) and again in `DeleteReplicationEntity` (line 243).

On its own this would only skip the remainder of one cycle. What makes it permanent is the caller. `processGroupState` in `internal/satellite/state/state_process.go` (line 465) returns as soon as `Replicate` errors, so it never reaches the block at line 482 that records what was accomplished:

```go
f.stateMap[index].State = newState
f.stateMap[index].Entities = FetchEntitiesFromState(newState)
```

Because `stateMap[index].Entities` is left unchanged, the next cycle's `GetChanges` compares the fetched state against the same old entity set and rebuilds an identical replicate list in an identical order. It appends while ranging over `newEntites` (line 199), which preserves the order of the state artifact, so the run fails at the same entity again.

`DeleteReplicationEntity` has the same structure and runs before replication (line 459), so a single undeletable artifact blocks replication for the entire group rather than just part of it.

## Current Behavior

A group's state lists artifacts A, B, C, D, E, and C is unreachable, for example deleted upstream, a per-repository credential failure, or a malformed manifest.

- Cycle 1: A and B replicate. C fails. D and E are never attempted. State is not saved.
- Cycle 2 (default `@every 00h00m10s`, `examples/config.example.json` line 9): identical list, identical order. A and B are re-checked, C fails again, D and E are never attempted.
- Steady state: D and E never replicate, and the satellite re-walks the same prefix every ten seconds indefinitely.

A workload that requires E fails to start, and the cause is C, which is unrelated to it. Whether an artifact ever lands depends on its position in the list relative to a broken one. Satellites run at edge and air-gapped sites where this is hard to observe: the only signal is a log line repeating every ten seconds, and the artifacts that never arrive are silent.

## Expected Behavior

A failure on one artifact does not prevent the remaining artifacts in the group from being attempted. Failures are collected and reported together, and artifacts that already replicated successfully are not re-walked from the start of the list on every subsequent cycle.

## Steps to Reproduce

1. Create a group whose state artifact lists several artifacts, ordered so a broken one is not last, for example `a:latest`, `broken:latest`, `c:latest`.
2. Make the middle artifact unresolvable, for example by deleting `broken:latest` from the source registry after the state artifact is published.
3. Start the satellite and let the replication loop run.
4. `a:latest` is replicated, `broken:latest` logs `Failed to fetch image descriptor`, and `c:latest` is never attempted. The sequence repeats every cycle and `c:latest` never appears in the local registry.

## Proposed Fix

Collect per-entity errors inside `Replicate` and continue through the list, returning an aggregate via `errors.Join` once the loop completes, and do the same in `DeleteReplicationEntity`. `processGroupState` can then decide whether partial progress should be committed to `stateMap` and persisted.

This pattern already exists in the repository. `ApplyCRIConfigs` in `internal/satellite/container_runtime/detect-cri.go` is documented as "Errors are collected per-CRI rather than failing on the first error", and the direct-delivery block inside `processGroupState` itself (line 471) logs its failures as warnings and continues. Replication is the outlier, including within the same function.

Harshitaakri:
- Relates to: #625

## Description

Both store implementations (`RegistryStore.Replicate` in `registry.go`, `OCIStore.Replicate` in `oci.go`) abort the entire replication run on the first per-entity error. This converts per-entity failures to warn-and-continue with `errors.Join`, so partial failure is reported to the caller while remaining artifacts are still replicated.

**Precedent:** `DirectDeliverer` (`state/direct_delivery.go`) already warn-and-continues per entity.

**Design choice:** continue + `errors.Join` so partial failure is still reported to the caller. `DirectDeliverer` currently logs and returns nil — question for maintainers: should all three paths converge on join-and-return or warn-and-nil?

**Scope note:** `Delete()` has the same abort pattern in both stores; deliberately deferred to keep this diff reviewable. On partial failure the caller (`processGroupState`) doesn't advance its recorded state, so the next cycle retries the full set — but already-replicated artifacts short-circuit on the stores' digest up-to-date check (`remote.Head` in `RegistryStore`, `Resolve` in `OCIStore`), so only failed entities do real work. Persisting partial progress in `processGroupState` (#625) would save those HEAD checks and is follow-up scope.

## Additional context

This PR does not use `Fixes #625` because the linked issue also requires the deletion path and `processGroupState` changes, which are out of scope here.

Note: `Delete` failure at `state_process.go:465` returns before `Replicate` runs — so with `Delete` deliberately unfixed here, one bad delete still blocks a group's entire replication. That's the sharpest justification for the Delete follow-up.

## Summary by CodeRabbit

* **Improvements**
  * Replication now processes multiple artifacts concurrently, improving throughput while maintaining controlled resource usage.

* **Bug Fixes**
  * Replication continues processing remaining artifacts when individual artifacts encounter validation, resolution, retrieval, or copy errors.
  * Successfully processed artifacts are preserved even when another artifact fails.
  * Replication reports encountered errors together with clearer artifact-specific details.
  * Context cancellation continues to stop replication immediately.

* **Tests**
  * Added coverage confirming valid artifacts are preserved when another artifact is unavailable.

### DIFF
--- a/internal/satellite/store/registry.go
+++ b/internal/satellite/store/registry.go
@@ -3,6 +3,7 @@ package store
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/http"
 
@@ -63,6 +64,7 @@ func (r *RegistryStore) Replicate(ctx context.Context, replicationEntities []Art
 		}
 	}
 
+	var errs []error
 	for _, entity := range replicationEntities {
 		// Check context cancellation before processing each image
 		select {
@@ -73,7 +75,9 @@ func (r *RegistryStore) Replicate(ctx context.Context, replicationEntities []Art
 		}
 
 		if err := entity.validate(); err != nil {
-			return err
+			log.Warn().Err(err).Str("entity", entity.Name).Msg("Skipping entity: validation failed")
+			errs = append(errs, fmt.Errorf("validate entity %s: %w", entity.Name, err))
+			continue
 		}
 
 		srcRef := r.source.reference(entity, entity.sourceIdentifier())
@@ -81,25 +85,31 @@ func (r *RegistryStore) Replicate(ctx context.Context, replicationEntities []Art
 
 		src, err := name.ParseReference(srcRef, nameOpts...)
 		if err != nil {
-			return fmt.Errorf("parse source ref %s: %w", srcRef, err)
+			log.Warn().Err(err).Str("entity", entity.Name).Str("ref", srcRef).Msg("Skipping entity: failed to parse source ref")
+			errs = append(errs, fmt.Errorf("parse source ref %s: %w", srcRef, err))
+			continue
 		}
 
 		dst, err := name.ParseReference(dstRef, nameOpts...)
 		if err != nil {
-			return fmt.Errorf("parse dest ref %s: %w", dstRef, err)
+			log.Warn().Err(err).Str("entity", entity.Name).Str("ref", dstRef).Msg("Skipping entity: failed to parse dest ref")
+			errs = append(errs, fmt.Errorf("parse dest ref %s: %w", dstRef, err))
+			continue
 		}
 
 		// Lazy fetch: only the manifest is downloaded, no layer data yet
 		desc, err := remote.Get(src, pullOpts...)
 		if err != nil {
-			log.Error().Msgf("Failed to fetch image descriptor: %v", err)
-			return err
+			log.Warn().Err(err).Str("entity", entity.Name).Str("ref", srcRef).Msg("Skipping entity: failed to fetch image descriptor")
+			errs = append(errs, fmt.Errorf("fetch image descriptor %s: %w", entity.Name, err))
+			continue
 		}
 
 		img, err := desc.Image()
 		if err != nil {
-			log.Error().Msgf("Failed to resolve image: %v", err)
-			return err
+			log.Warn().Err(err).Str("entity", entity.Name).Msg("Skipping entity: failed to resolve image")
+			errs = append(errs, fmt.Errorf("resolve image %s: %w", entity.Name, err))
+			continue
 		}
 
 		// Lazy OCI conversion, no data materialized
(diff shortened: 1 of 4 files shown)


internal/satellite/store/registry.go
```diff
@@ -130,13 +144,14 @@ func (r *RegistryStore) Replicate(ctx context.Context, replicationEntities []Art
...
-			return err
+			log.Warn().Err(err).Str("entity", entity.Name).Msg("Skipping entity: failed to replicate image")
+			errs = append(errs, fmt.Errorf("replicate image %s: %w", entity.Name, err))
+			continue
 		}
 		log.Info().Msgf("Image %s replicated successfully", entity.Name)
 	}
 
-	return nil
+	return errors.Join(errs...)
```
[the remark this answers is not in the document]
Harshitaakri: Good catch — this is the same concern flagged in the PR description under "Open question: join vs nil."

The current behavior: `errors.Join(errs...)` returns non-nil when any entity fails → `state_process.go:471` early-returns → `DirectDeliverer.Deliver` at `:477` never runs for the successfully-replicated artifacts in that cycle.

This is intentional for now — the caller's next cycle retries the full set, and each store's digest check (`remote.Head` / `target.Resolve`) short-circuits artifacts that already landed, so they're not re-pulled. But the delivery delay for the successful subset is real.

The fix belongs in `processGroupState` (separating replication errors from the deliver gate), tracked in #625. Changing the return here would hide partial failures from the caller, which is worse.

internal/satellite/store/registry.go
```diff
@@ -63,6 +64,7 @@ func (r *RegistryStore) Replicate(ctx context.Context, replicationEntities []Art
 		}
 	}
 
+	var errs []error
 	for _, entity := range replicationEntities {
```
bupd, line 96: this line should be fixed with queue. instead of a for loop and should be dispatched to goroutines.

bupd (changes requested):
Thanks for your contribution @Harshitaakri 

I have suggested changes please update

internal/satellite/store/registry.go
Harshitaakri: Good call!
Refactored in `97c3f37`. The replication loop now dispatches entities through a channel to a pool of 5 concurrent worker goroutines. Each worker picks entities off the queue and calls `replicateEntity `independently. Errors are collected via a mutex-protected slice and joined at the end, so partial failures still report correctly. Context cancellation stops dispatching and drains in-flight workers.

Note: `OCIStore.Replicate` is kept sequential intentionally. It holds a mutex because the OCI layout store isn't safe for concurrent writes.

PR #648, opened 2026-08-22
feat: Add generic OCI store interface by replacing Zot with ORAS

### LINKED ISSUE #646: feat: Add generic OCI store interface by replacing Zot with ORAS
# Overview

Harbor Satellite currently depends on Zot as the default storage backend for caching and replicating OCI images. This dependency introduces an embedded registry where the core requirement is only to persist and transfer OCI content.

Replace Zot with an ORAS-based storage layer that stores artifacts directly in a local OCI image layout. The design should also introduce a backend-agnostic store interface so replication workflows can operate consistently with either local OCI storage or a user-provided remote registry.

## Problem

Using Zot as an internal storage layer adds unnecessary operational and architectural complexity:

- Satellite must manage an embedded registry process, configuration, lifecycle, and health solely to retain replicated artifacts.
- Storage behavior is coupled to Zot-specific registry semantics instead of the OCI content model.
- Local caching requires a running registry service even though an OCI-compliant filesystem layout is sufficient.
- Replication logic does not have a clear abstraction separating source artifact references, destination repositories, authentication, and storage implementation.
- Source and destination repository paths can become coupled, causing source namespaces to leak into destination references during remote-to-remote replication.
- Registry repositories may be empty or contain multiple path segments, so references cannot assume a fixed `<project>/<image>` structure.
- Supporting alternative storage backends would require replication workflows to understand backend-specific details.

These limitations make the current implementation harder to maintain and restrict future work described by the transparent OCI registry proxy ADR.

## Solution

Introduce a generic OCI store abstraction that exposes only the operations required by Satellite, including artifact replication and deletion.

Provide two implementations:

- A local OCI store backed by `oras-go` and its `content/oci` package. It should pull content from a remote registry and persist manifests, configurations, and layers in a standards-compliant OCI image layout on the local filesystem.
- A registry store for BYO registry deployments. It should preserve the existing remote-registry workflow while supporting remote-to-remote OCI artifact replication.

Model source and destination registry settings independently, including endpoint, repository path, credentials, transport security, and plain HTTP behavior. Artifact metadata should identify the mandatory image name and optional repository path, tag, or digest without assuming registry-specific project structures.

Reference construction must correctly support:

- Registries with no repository prefix.
- Single-level and multi-level repository paths.
- Tag-based and digest-based source references.
- Independent source and destination namespaces.
- Destination tags that differ from the source digest.
- Authentication scoped to the corresponding source or destination registry.

The resulting design should remove Zot from Satellite’s storage path, retain the current replication flow, keep existing Crane usage where still required, and provide an extensible boundary for future local or remote OCI storage implementations.

vg006:
# Overview

- Fixes: #646

## Description

- Replaces Zot with an ORAS-backed local OCI image-layout store.
- Renames the satellite registry package to `store` and introduces backend-agnostic replication and deletion interfaces.
- Adds local OCI and BYO remote-registry store implementations.
- Separates source and destination registry endpoints, repository paths, credentials, and artifact references.
- Supports empty or multi-level repository paths and tag- or digest-based replication.
- Updates satellite state processing, configuration, dependencies, tests, and deployment examples for the new storage model.
- Adds a test utility for validating remote-to-local and remote-to-remote replication.

## Additional context

Crane remains in use where currently required. This PR focuses only on removing Zot and establishing the ORAS-based storage abstraction; transparent proxy behavior from the ADR is outside its scope.

<a href="https://cubic.dev/pr/container-registry/harbor-satellite/pull/648?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bupd (approved):
/lgtm

Issue #665, opened 2026-09-14
satellite peer to peer


bupd:
<img width="1928" height="973" alt="Image" src="https://github.com/user-attachments/assets/dcbd44b4-2930-4d2c-a1f9-d8330ef41f41" />

bupd:
https://excalidraw.com/#json=A5lMTQ-Bn0iMPGMFnd20g,v0MDcUqwEjZDy2qP3z8gZg - excalidraw

PR #671, opened 2026-09-21
Docs/adr 0010 peer to peer distribution

cotishq:
- Related: #542 
- Related: #665 

## Description

Draft **proposed** ADR for opt-in, air-gapped copy of desired-state OCI artifacts between trusted Satellites. Docs only - no runtime change.

Peer `/v2/` is ADR-0009 **replica mode** on `PROXY_PORT` (#669), not a second facade. Copy stays in the existing desired-state loop (`oras.Copy`, canonical Harbor refs). Harbor is fallback only when no peer delivered a complete artifact.

This revision also locks:
- same-group peer selection by default; `REACHOUT_SATS=global` is the override
- union of static URLs and optional `gc_peers` (empty is valid; no GC peer APIs in this PR)
- cancel in-flight pulls only when a peer returns the **complete** digest-verified graph
- GC owns peer health when present; unreachable listed peer: retry once, log, fail - no silent skip / local blacklist

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for opt-in peer-to-peer OCI artifact distribution between trusted Satellites on isolated networks.
  * Documented trusted peer discovery, group filtering, roster handling, credential and HTTPS security requirements, and configuration options.
  * Clarified local replica misses, proxy-only upstream filling, degraded-state behavior, graph admission, and quarantine handling.
  * Added the new architecture decision to the documentation index.


cotishq:
Ready for a look,

I tightened the architecture wording (bootstrap desired-state file, unique peer list, content policy) and added an overview diagram so the peer-copy flow is easier to follow as discussed in the sync.

happy to adjust, if something feels off,

CC: @bupd @vg006

